Issue #89, opened 2021-01-03
Linux Support


luciusDXL:
The Force Engine should compile and run on common Linux Distros using the CMake build system.

gilmorem560:
Not sure when this will be getting attention/be more relevant, but with some tweaking, I was able to produce a Linux build:

![image](https://user-images.githubusercontent.com/36053668/143734359-04d71e0b-0b61-413c-aedd-c0c0d67db320.png)

I'm still working on actually getting the game to load, and cut a few corners in the threading stuff up under the MIDI player for now, but things seem to mostly just work as far as initially launching and loading the menu.

One thing I did notice throughout many files is the use of strcpy and memcpy without any inclusion of the cstring header, so I had to add that in quite a few places before g++ would build it. Additionally, some files using INT_MAX among other symbols required climits. The pthread implementation of the threading object is incomplete, but will build in at least. Finally, I needed to provide POSIX equivalents of many of the fileutil methods. I'll respond here with more info as I come across it, and likely will be raising a pull request once I've got the changes tidied up and tested.

Note, I am not intending on doing any build-system stuff, at least not yet, but just intend to contribute modifications to the source code itself that allow the files to build happily using g++ on an X11 POSIX system.

gilmorem560:
And now I've actually got it launching the game:

![image](https://user-images.githubusercontent.com/36053668/143787966-74361023-ffe0-486a-ae7c-7a9b4ae1e5b8.png)

No audio yet, but I can run around and shoot things. I haven't explored outside the first room yet, just getting a feel for how well things are working thus far. The mouse cursor does not appear to be captured in the SDL window, I don't know if this is the experience on other platforms, and I don't have access to a decent Windows box to see how that build behaves. Currently the mouse can and *will* interact with items besides just the SDL window, making actual gameplay outside of fullscreen mode incredibly difficult as clicking anywhere outside the immediate area of the window unfocuses...

luciusDXL:
On Windows, the mouse is locked to the window while in gameplay.

gilmorem560:
Reporting back that I've done some further testing but been busy with other things. I am able to go from level to level it seems, so I think things are behaving themselves pretty well all things considered. This weekend I'm going to see about massaging the rest of the necessary bits into enough shape to submit a succinct PR. The Mac stuff ala #90 will probably be a bit more back burner, but as time allows I intend on continuing to futz with that as well. Still don't have audio on either, but haven't been focused on that.

[...]

PR #99, opened 2021-11-28
Fix Simple Missing/Discrepant Includes

gilmorem560:
This PR just touches files that needed some header modifications to build on Linux, this does *not* produce a fully functional build. Supports #89


luciusDXL:
It looks good.

PR #110, opened 2022-01-12
Separating weapon wave from head wave...

ExtraDrowsy:
This is to address issue `106` for head waving/bobbing. Based on the discussion that already took place:
* The head and weapon wave/bob together.
* In the original game (and TFE) both can be disabled together using the F6 key.
* A request was made to allow the head wave to be disabled but keep the weapon wave on.

To do this I added a new checkbox type setting to the "Hud" settings that controls weapon waving independently/separately from head waving.

At first I considered making this a hotkey like "head wave" is but decided it was not ideal for the following reasons:
* Separate weapon wave functionality are not part of the original game, the hotkeys are.
* The hotkeys appear to rely on original game files (I might be wrong), like Strings in `.msg` files that render as on/off messages to the screen when the `F6` hotkey is pressed. Those Strings belong to the game not the source engine/port and I think original/base game files should always be utilized in a read-only manner which means rather than running through the same code that looks through the parsed `.msg` files, we'd have to have something that sources the Strings from literals in code or some other file.
* Having two hotkeys for waving seemed a bit excessive.


TheForceEngine/TFE_DarkForces/player.cpp
```diff
@@ -1830,69 +1830,76 @@ namespace TFE_DarkForces
 			//s_282344 = 34 - dH;
 		}
```
ExtraDrowsy, line 1832: This diff below looks big but all of the lines from this point to `:1844` are unchanged except for the following:
* I removed the `if` block that starts on line `:1836` and thus all of the code within that block down to `:1885` was indented left-ward (hence why the diff looks large). I removed it because we now need to run these calculations regardless of whether head waving is on since these are used for weapon waving.
* Any assignments to `s_headwaveVerticalOffset` in this range of code now assign to a new local variable `headwaveVerticalOffset` (see `:1835`). I essentially avoid modifying `s_headwaveVerticalOffset` until I know the head waving is on and needs to be applied since modifying `s_headwaveVerticalOffset` alters how player projectiles (like lasers from the blaster) are calculated.

TheForceEngine/TFE_DarkForces/player.cpp
```diff
@@ -1830,69 +1830,76 @@ namespace TFE_DarkForces
...
+				fixed16_16 speed = externSpd & (~0xffff);
+				speed |= (s_curTick & 0xffff);
+				// Then multiply by 16 and take the cosine - this is meant to be a small modification to the weapon motion
+				// (note that the base multiplier is 128).
+				sinCosFixed(speed << 4, &s_wpnSin, &s_wpnCos);
+				// Modify the headwave motion.
+				headwaveVerticalOffset += mul16(s_wpnCos, PLAYER_HEADWAVE_VERT_WATER_SPD);
 			}
 		}
-		setCameraOffset(0, s_headwaveVerticalOffset, 0);
```
ExtraDrowsy, line 1887: The code below is where the condition was moved to and what ultimately separates the head waving and weapon waving, as well as applying the new setting.

luciusDXL:
Hello, thanks for all the work with this. I have been focused on the GPU Renderer and version 1.0 but I will take a look at what you did here.

PR #117, opened 2022-06-18
Move from critical section to SRWLocks

santagada:
SRWLocks are smaller (8b vs 40b for critical section) and faster to lock/unlock


luciusDXL:
Thanks for the PR. I will test it once some synchronization cleanup is done in the audio systems.

santagada:
yeah I was trying to use these mutexes in that part of the code (right now it is just using some macros that also use critical sections internally) but didn't finish it.

PR #120, opened 2022-10-02
Replace DevIL with stb_image

winterheart:
Replace DevIL library with embedded stb_image and stb_image_write functions from https://github.com/nothings/stb (under MIT license). This reduces libraries dependency footprint, as TFE requires only PNG loading/writing.


twolife:
is there any interest for this from the project leader ?
I have spend a little bit of time on this & can share an updated patch if there is any chance for it to be merged

PR #154, opened 2022-12-20
add command-line support for changing the midi device

codengine:
This PR adds a command-line parameter which allows the selection of a midi device other than the default.

The workflow with this solution is as follows:

- Start the engine
- All available midi devices with their port number are logged
- Get the desired port number from the logs
- Restart the engine using the command line parameter "--midi X", whereas X is the port number

I know that it would be better to have it configurable in the settings, but I only have very basic C++ knowledge :)

### DIFF
--- a/TheForceEngine/TFE_Audio/midiDevice.cpp
+++ b/TheForceEngine/TFE_Audio/midiDevice.cpp
@@ -29,6 +29,16 @@ namespace TFE_MidiDevice
 		s_midiout->setErrorCallback(midiErrorCallback);
 		s_openPort = -1;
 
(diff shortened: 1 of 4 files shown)


luciusDXL:
Sorry, I'm not planning on using this approach. I am going to just add the audio devices to the Sound menu and make them selectable at runtime.

luciusDXL:
The work is already in progress for the next build, but thanks for trying to help. :)

codengine:
I'm not sad at all, I'm happy that soon everyone can get it. Maybe that was just the spark that was needed :)
And btw. thanks for your great work!

PR #161, opened 2022-12-22
Basic Linux Port for 1.0

mlauss:
This is a basic linux port, tested with gcc-13 on x64.  It contains a few commits to fix issues which were found during testing; most of them are compile fixes for gcc which should also work on windows.
Thanks to Matthew Gilmore for the initial version, I took some code from your version as well.

The game is fully completeable, with software and hardware rendering, and with digital sound and midi music using e.g. fluidsynth. Mods work as well (only tested the "tie defender base" mod though).  There's still issues with reloading a savegame during play.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ Build/
 TheForceEngine/bug_reports/
 WDFuseCommonFiles/
 x64/Debug/
+
+# Linux
+TheForceEngine/tfelnx
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,286 @@
+CC	=	gcc
+CXX	=	g++
+LD	=	g++
+# -O2 and higher cause random segfaults.
+CXXFLAGS=	-O1 -ggdb3 -gdwarf-5 -pipe -I./TheForceEngine -I/usr/include/SDL2 -I/usr/include/rtaudio -I/usr/include/rtmidi
+LDFLAGS	=	-lpthread -lGL -lGLEW -lSDL2 -lIL -lILU -lrtaudio -lrtmidi -lasound
+
+ARCH	=	TheForceEngine/TFE_Archive/archive.o		\
+		TheForceEngine/TFE_Archive/gobArchive.o		\
+		TheForceEngine/TFE_Archive/gobMemoryArchive.o	\
+		TheForceEngine/TFE_Archive/labArchive.o		\
+		TheForceEngine/TFE_Archive/lfdArchive.o		\
+		TheForceEngine/TFE_Archive/zipArchive.o		\
+		TheForceEngine/TFE_Archive/zip/zip.o
+
+ASS	=	TheForceEngine/TFE_Asset/assetSystem.o		\
+		TheForceEngine/TFE_Asset/colormapAsset.o	\
+		TheForceEngine/TFE_Asset/dfKeywords.o		\
+		TheForceEngine/TFE_Asset/fontAsset.o		\
+		TheForceEngine/TFE_Asset/gameMessages.o		\
+		TheForceEngine/TFE_Asset/gifWriter.o		\
+		TheForceEngine/TFE_Asset/gmidAsset.o		\
+		TheForceEngine/TFE_Asset/imageAsset.o		\
+		TheForceEngine/TFE_Asset/levelList.o		\
+		TheForceEngine/TFE_Asset/levelObjectsAsset.o	\
+		TheForceEngine/TFE_Asset/modelAsset.o		\
+		TheForceEngine/TFE_Asset/modelAsset_jedi.o	\
+		TheForceEngine/TFE_Asset/paletteAsset.o		\
+		TheForceEngine/TFE_Asset/spriteAsset.o		\
+		TheForceEngine/TFE_Asset/spriteAsset_Jedi.o	\
+		TheForceEngine/TFE_Asset/textureAsset.o		\
+		TheForceEngine/TFE_Asset/vocAsset.o		\
+		TheForceEngine/TFE_Asset/vueAsset.o
+
+AUD	=	TheForceEngine/TFE_Audio/audioDevice.o		\
+		TheForceEngine/TFE_Audio/audioSystem.o		\
+		TheForceEngine/TFE_Audio/midiDevice.o		\
+		TheForceEngine/TFE_Audio/midiPlayer.o
+
+DF	=	TheForceEngine/TFE_DarkForces/Actor/actor.o			\
+		TheForceEngine/TFE_DarkForces/Actor/actorSerialization.o	\
+		TheForceEngine/TFE_DarkForces/Actor/animTables.o		\
+		TheForceEngine/TFE_DarkForces/Actor/bobaFett.o			\
+		TheForceEngine/TFE_DarkForces/Actor/dragon.o			\
+		TheForceEngine/TFE_DarkForces/Actor/enemies.o			\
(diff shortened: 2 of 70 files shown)


luciusDXL:
Thank you for your efforts. I will go through this when I can, though it might be after the holidays.

luciusDXL:
I plan on tackling Linux support after the holidays, so I will go through this and pick it up then. Thanks again for everyone's efforts.

spacekomet:
Nice work, @mlauss .

I had no problem compiling it and music also plays with the software synthesizer. I didn't notice any issues so far, with exception of green colors caused by my Intel HD4000 I guess, even with the latest mesa driver, which can however be fixed with changing the values in fraq shaders files. 

But I get some issue with the mod.
I've tried playing Dark Tide 1 mode, and there I've noticed I get agents and mission list in the agent menu from the main game and not from the mod. If I then choose mission one (Secret Base) the mod starts.  
Strangely enough, correct agent menu is displayed using older linux port (however, mod was not even playable with that one, because it just froze when trying to pick first shield pack at the beginning, probably not all files were loaded due the name casings). 

One more thing, when exiting the mod (via the game or TFE menu), I get the segmentation fault. 
And during the gameplay I get a lot  iMuse errors:
[DEBUG] [Error : iMuse] ERR: Illegal chunk in sound 984383974039136...

mlauss:
Hi,

I don't see your problem with the Dark Tide 1 Mod -- in the Agent Menu,
after a lot of cutscenes, I see a single entry
"Wasteland Survivor", and it plays just fine.   I suggest you start with a
fresh installation of DF, without any
changed file cases, etc. That's what I do tests on.
I know about the segfaults, I'll try to debug it with valgrind in the
coming days.

Manuel

HandMeBacon:
I can confirm that this is working in Linux Mint. Only problem I've had is my desktop will crash if I alt+tab while the game is running. Weirdly, Mint just kicks me back to the login screen after closing my apps.

gilmorem560:
Wanted to throw in that I've also gone through and done a fresh 1.0 port using upstream, I'll take a look at this later and see how much parity there still is. One thing I did differently was extract ImGui and use a copy of that from the package repos. Worth noting that introduces a few caveats:

- ImGui markdown is provided by a third-party library: https://github.com/juliettef/imgui_markdown

- The above (or perhaps Debian's libimgui-dev package) is then dependent on some stb libraries, libstb-dev.

One curious matter is TheForceEngine uses the method linked here https://github.com/ocornut/imgui/issues/1167#issuecomment-306652547

However, I can't seem to find this anywhere in the copy of ImGui in Debian's package system. That could point to Debian having an older version or otherwise this being non-standard, but just bringing this up as this does have implications for portability.

Is there a reason all the dependencies are static by the way? As visible in the Makefile, I kinda stripped all those out of the Linux build, opting to shared libraries on the host machine instead, which at the time at least appeared to work. If there are any specific customizations to these external libraries, those would present portability blocks as well.

mlauss:
> Wanted to throw in that I've also gone through and done a fresh 1.0 port using upstream, I'll take a look at this later and see how much parity there still is. One thing I did differently was extract ImGui and use a copy of that from the package repos. Worth noting that introduces a few caveats:
> 
>     * ImGui markdown is provided by a third-party library: https://github.com/juliettef/imgui_markdown
> 
>     * The above (or perhaps Debian's libimgui-dev package) is then dependent on some stb libraries, libstb-dev.

Maybe we can use ImGui freetype support, that should work with almost any distro out of the box.
But isn't ImGui intended to be used like it is in TFE?  I can't seem to find an "official" way to build shared library.

gilmorem560:
Maybe you're right, the libimgui-dev package only seems to include a .a static library. Can't say I know much about the specifics, was just combing for what dependencies could be offloaded to shared libraries or distro packages. Looks like Debian, Red Hat, and Arch all have their own packages for this. I'm not sure what "standard" practice is out there in Linux-world with imgui, but the package managers seem to suggest there are at least some attempts to distribute this universally rather than it just being an internal dependency on stuff, but what with the .a library only, not sure what they're actually achieving.

In any case, the imgui_markdown project linked has to be explicitly pulled in as it isn't part of upstream nor is it in any package managers. If the project is specifically dependent on a modified local copy of imgui, of course, this all becomes moot, but if the intent is to consume imgui as it stands out there in the world, then different distros may have different opinions on what revision they ship, custom changes, etc. Thus is the landscape in Linux-world...

mlauss:
> Edit: Just to add, I've got quite a bit of pruning to do on my fork of 1.0 before it's ready to even consider merging, I don't think I'm actually going to do all of that hubbub since this PR exists. Instead, once I get a working copy, I'm going to three-way compare all of them and anything that seems to have improved in my re-port I'll try and get in a place that can merge together with all of this.

How's stability? The code compiles now fine with gcc and clang, but any optimizations higher than -O1 generate very crashy binaries on either compiler.  The memory allocator code seems most fragile. I've started trying to eliminate a lot of warnings, and sometimes just moving variables around causes memory exhaustion ...

Gerwin2k:
Hey, mlauss, past weekend I took your TFE Linux 1.0 repo and build it for Windows XP x86 (=32 bit) with MinGW32. Just for curiosity. Thanks.
I had to add some workarounds. Indeed any optimize for speed option made the game crash (-O, -O1, -O2, -O3), usually when a round is fired by player or enemy. Optimization for size -Os did work fine, so used that instead. No crashes since.

(files on my website http://www.gb-homepage.nl/ in the back ports section)

Gerwin2k:
Maybe related to this?
"optimized code gives strange floating point results"
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=323

spacekomet:
> Hi, I don't see your problem with the Dark Tide 1 Mod -- in the Agent Menu, after a lot of cutscenes, I see a single entry "Wasteland Survivor", and it plays just fine. I suggest you start with a fresh installation of DF, without any changed file cases, etc. That's what I do tests on. I know about the segfaults, I'll try to debug it with valgrind in the coming days. Manuel

Clean installation helped, thanks for the hint. I suspect the issue was german patch installed over the english version of the game. which changed the mission names. 

However, I get dumps, when I try to load (quick or normal) a saved game in the mod (Dark Tide1, didn't try with others). 
TFE is compiled up to the latest clang commit [a2cbc1d8](https://github.com/mlauss/TheForceEngine/commit/a2cbc1d87ae4829902a62e18cb96b6b366d87a0d) . 
Steps:
1. Start the mod and the 1st level
2. Save the game, with ALT+F5 (quick save) or going into TFE menu and save it there. 
4. Load the game with ALT+F9 or via the TFE menu. 
4. Game dumps. 

Loading otherwise works  fine in  the normal game. 
I sure hope I'm not the only one experiencing this :).

mlauss:
I can reproduce your save-then-load crash (startup-then-load does not crash), it's caused by the archive freeing running into some member memory corruption, specifically the first Archive* in s_globalArchives contains garbage already.  With -O2 or higher, these types of crashes become more widespread, i.e the crash when the grate lifts at the start of mission 1 is also caused by corruption of an "Allocator" structure when the INF code deletes this elevator.

mlauss:
I just pushed a small patch which practically fixes almost all runtime crashes when building the source with GCC and O2/O3 and clang at O1 or higher.  I was able to complete SECBASE with all secrets just fine. LTO does not yet work, but I'm looking into it.

PR #170, opened 2022-12-28
Explicit template specialization is not supported under GCC yet

philberty:
This patch removes the explicit template specialization by simply naming these methods and calling them explicitly as the caller is not templated so it doesnt affect the code path.

For more info see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85282

Note I have a fork on github starting to add CMakeLists and trying to build on linux over here: https://github.com/philberty/theforceengine


luciusDXL:
This change was handled in a different PR.

PR #171, opened 2022-12-29
Added MIDI Device Selection in Sound Settings

sithmartyr:
Added the ability to select a different MIDI output device, so that you can use a Roland SoundCanvas.

https://youtu.be/zA4txQuixS0

image

Issue #159, opened 2022-12-22
Feature request: 3D audio (5.1 / 7.1 / "surround sound")


seascape:
Congratulations on shipping 1.0. A day long awaited!

I have an audio-related suggestion. GZDoom has a wonderful implementation of the [OpenAL Soft audio middleware](https://www.openal-soft.org/).  It can position sounds in 3D space and convincingly simulate those placements audibly through headphones with HRTF (head-related transfer function) effects. 

A similar feature would be an obvious fit for an FPS like Dark Forces, greatly enhancing its soundscape.

QmwJlHuSg9pa:
> Virtualized headphone 3D audio via HRTFs is also common in VR, modern games consoles like PS5, etc.
> 
> A similar feature would be an obvious fit for an FPS like Dark Forces, greatly enhancing its soundscape.

In terms of lineage, all of the sequels implement 3D audio -- either through the A3D (JKDF2) or OpenAL (Outcast, Academy) APIs (for best compatibility with modern machines, one can utilise drop-in _OpenAL Soft_ libraries in all cases). Also, modern projects like OpenJKDF2 and OpenJK use OpenAL Soft.

Indeed, this would be a "nice to have" feature.

luciusDXL:
TFE was aiming for accuracy in terms of sound as well as other aspects. For that reason, the game code and iMuse are used for sound mixing. There are plans for improved sound in the future and support for things like this, but I want to make sure the option exists to keep the original sound mixing if desired as well. Anyway, definitely an area for future work. :)

PR #176, opened 2022-12-30
Updated Linux Support for 1.01

mlauss:
Linux Port for 1.01.  With these changes, the engine runs on 32bit and 64bit linux (tested x64 and MIPS32), Dark Forces is fully completable (tested english and german versions), with OpenGL rendering, Sound and Music (tested with fluidsynth).  Mods are working as well (tested with Tie Defender Base)
Loading savegames sometimes crashes, I'm still looking into that.

The first half of patches fix general issues found during testing, the other half is gcc/clang/linux-specific changes.
Big thanks to Matthew Gilmore, I took a few patches and other changes from your initial port.

Source build tested with gcc-11, gcc-12, gcc-13 and clang-15.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ Build/
 TheForceEngine/bug_reports/
 WDFuseCommonFiles/
 x64/Debug/
+
+# Linux
+TheForceEngine/tfelnx
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,293 @@
+#
+# Makefile to build TFE on Linux.
+#
+CC	=	gcc
+CXX	=	g++
+LD	=	g++
+CXXFLAGS=	-O2 -pipe -I./TheForceEngine -I/usr/include/SDL2 -I/usr/include/rtaudio -I/usr/include/rtmidi
+LDFLAGS	=	-lpthread -lOpenGL -lGLEW -lSDL2 -lIL -lILU -lrtaudio -lrtmidi -lasound
+
+ARCH	=	TheForceEngine/TFE_Archive/archive.o		\
+		TheForceEngine/TFE_Archive/gobArchive.o		\
+		TheForceEngine/TFE_Archive/gobMemoryArchive.o	\
+		TheForceEngine/TFE_Archive/labArchive.o		\
+		TheForceEngine/TFE_Archive/lfdArchive.o		\
+		TheForceEngine/TFE_Archive/zipArchive.o		\
+		TheForceEngine/TFE_Archive/zip/zip.o
+
+ASS	=	TheForceEngine/TFE_Asset/assetSystem.o		\
+		TheForceEngine/TFE_Asset/colormapAsset.o	\
+		TheForceEngine/TFE_Asset/dfKeywords.o		\
+		TheForceEngine/TFE_Asset/fontAsset.o		\
+		TheForceEngine/TFE_Asset/gameMessages.o		\
+		TheForceEngine/TFE_Asset/gifWriter.o		\
+		TheForceEngine/TFE_Asset/gmidAsset.o		\
+		TheForceEngine/TFE_Asset/imageAsset.o		\
+		TheForceEngine/TFE_Asset/levelList.o		\
+		TheForceEngine/TFE_Asset/levelObjectsAsset.o	\
+		TheForceEngine/TFE_Asset/modelAsset.o		\
+		TheForceEngine/TFE_Asset/modelAsset_jedi.o	\
+		TheForceEngine/TFE_Asset/paletteAsset.o		\
+		TheForceEngine/TFE_Asset/spriteAsset.o		\
+		TheForceEngine/TFE_Asset/spriteAsset_Jedi.o	\
+		TheForceEngine/TFE_Asset/textureAsset.o		\
+		TheForceEngine/TFE_Asset/vocAsset.o		\
+		TheForceEngine/TFE_Asset/vueAsset.o
+
+AUD	=	TheForceEngine/TFE_Audio/audioDevice.o		\
+		TheForceEngine/TFE_Audio/audioSystem.o		\
+		TheForceEngine/TFE_Audio/midiDevice.o		\
+		TheForceEngine/TFE_Audio/midiPlayer.o
+
+DF	=	TheForceEngine/TFE_DarkForces/Actor/actor.o			\
+		TheForceEngine/TFE_DarkForces/Actor/actorSerialization.o	\
+		TheForceEngine/TFE_DarkForces/Actor/animTables.o		\
+		TheForceEngine/TFE_DarkForces/Actor/bobaFett.o			\
+		TheForceEngine/TFE_DarkForces/Actor/dragon.o			\
+		TheForceEngine/TFE_DarkForces/Actor/enemies.o			\
+		TheForceEngine/TFE_DarkForces/Actor/exploders.o			\
+		TheForceEngine/TFE_DarkForces/Actor/flyers.o			\
+		TheForceEngine/TFE_DarkForces/Actor/mousebot.o			\
+		TheForceEngine/TFE_DarkForces/Actor/phaseOne.o			\
+		TheForceEngine/TFE_DarkForces/Actor/phaseThree.o		\
+		TheForceEngine/TFE_DarkForces/Actor/phaseTwo.o			\
+		TheForceEngine/TFE_DarkForces/Actor/scenery.o			\
+		TheForceEngine/TFE_DarkForces/Actor/sewer.o			\
+		TheForceEngine/TFE_DarkForces/Actor/troopers.o			\
+		TheForceEngine/TFE_DarkForces/Actor/turret.o			\
+		TheForceEngine/TFE_DarkForces/Actor/welder.o			\
+		TheForceEngine/TFE_DarkForces/GameUI/agentMenu.o		\
+		TheForceEngine/TFE_DarkForces/GameUI/delt.o			\
+		TheForceEngine/TFE_DarkForces/GameUI/editBox.o			\
+		TheForceEngine/TFE_DarkForces/GameUI/escapeMenu.o		\
+		TheForceEngine/TFE_DarkForces/GameUI/menu.o			\
+		TheForceEngine/TFE_DarkForces/GameUI/missionBriefing.o		\
+		TheForceEngine/TFE_DarkForces/GameUI/pda.o			\
+		TheForceEngine/TFE_DarkForces/GameUI/uiDraw.o			\
+		TheForceEngine/TFE_DarkForces/Landru/cutscene.o			\
+		TheForceEngine/TFE_DarkForces/Landru/cutsceneList.o		\
+		TheForceEngine/TFE_DarkForces/Landru/cutscene_film.o		\
+		TheForceEngine/TFE_DarkForces/Landru/cutscene_player.o		\
+		TheForceEngine/TFE_DarkForces/Landru/lactor.o			\
+		TheForceEngine/TFE_DarkForces/Landru/lactorAnim.o		\
+		TheForceEngine/TFE_DarkForces/Landru/lactorCust.o		\
+		TheForceEngine/TFE_DarkForces/Landru/lactorDelt.o		\
+		TheForceEngine/TFE_DarkForces/Landru/lcanvas.o			\
+		TheForceEngine/TFE_DarkForces/Landru/ldraw.o			\
+		TheForceEngine/TFE_DarkForces/Landru/lfade.o			\
+		TheForceEngine/TFE_DarkForces/Landru/lfont.o			\
+		TheForceEngine/TFE_DarkForces/Landru/lmusic.o			\
+		TheForceEngine/TFE_DarkForces/Landru/lpalette.o			\
+		TheForceEngine/TFE_DarkForces/Landru/lrect.o			\
+		TheForceEngine/TFE_DarkForces/Landru/lsound.o			\
+		TheForceEngine/TFE_DarkForces/Landru/lsystem.o			\
+		TheForceEngine/TFE_DarkForces/Landru/ltimer.o			\
+		TheForceEngine/TFE_DarkForces/Landru/lview.o			\
+		TheForceEngine/TFE_DarkForces/Landru/textCrawl.o		\
+		TheForceEngine/TFE_DarkForces/agent.o				\
+		TheForceEngine/TFE_DarkForces/animLogic.o			\
+		TheForceEngine/TFE_DarkForces/automap.o				\
+		TheForceEngine/TFE_DarkForces/briefingList.o			\
+		TheForceEngine/TFE_DarkForces/cheats.o				\
+		TheForceEngine/TFE_DarkForces/config.o				\
+		TheForceEngine/TFE_DarkForces/darkForcesMain.o			\
+		TheForceEngine/TFE_DarkForces/gameMusic.o			\
+		TheForceEngine/TFE_DarkForces/gameMessage.o			\
+		TheForceEngine/TFE_DarkForces/generator.o			\
+		TheForceEngine/TFE_DarkForces/hitEffect.o			\
+		TheForceEngine/TFE_DarkForces/hud.o				\
+		TheForceEngine/TFE_DarkForces/item.o				\
+		TheForceEngine/TFE_DarkForces/logic.o				\
+		TheForceEngine/TFE_DarkForces/mission.o				\
+		TheForceEngine/TFE_DarkForces/pickup.o				\
+		TheForceEngine/TFE_DarkForces/player.o				\
+		TheForceEngine/TFE_DarkForces/playerCollision.o			\
+		TheForceEngine/TFE_DarkForces/projectile.o			\
+		TheForceEngine/TFE_DarkForces/random.o				\
+		TheForceEngine/TFE_DarkForces/sound.o				\
+		TheForceEngine/TFE_DarkForces/time.o				\
+		TheForceEngine/TFE_DarkForces/updateLogic.o			\
+		TheForceEngine/TFE_DarkForces/util.o				\
+		TheForceEngine/TFE_DarkForces/vueLogic.o			\
+		TheForceEngine/TFE_DarkForces/weapon.o				\
+		TheForceEngine/TFE_DarkForces/weaponFireFunc.o
+
+FS	=	TheForceEngine/TFE_FileSystem/filestream-posix.o	\
+		TheForceEngine/TFE_FileSystem/fileutil-posix.o		\
+		TheForceEngine/TFE_FileSystem/filewriterAsync.o		\
+		TheForceEngine/TFE_FileSystem/memorystream.o		\
+		TheForceEngine/TFE_FileSystem/paths-posix.o
+
+FE	=	TheForceEngine/TFE_FrontEndUI/console.o			\
+		TheForceEngine/TFE_FrontEndUI/editorTexture.o		\
+		TheForceEngine/TFE_FrontEndUI/frontEndUi.o		\
+		TheForceEngine/TFE_FrontEndUI/modLoader.o		\
+		TheForceEngine/TFE_FrontEndUI/profilerView.o
+
+GAME	=	TheForceEngine/TFE_Game/igame.o				\
+		TheForceEngine/TFE_Game/reticle.o			\
+		TheForceEngine/TFE_Game/saveSystem.o 
+
+INPUT	=	TheForceEngine/TFE_Input/input.o			\
+		TheForceEngine/TFE_Input/inputMapping.o
+
+JEDI	=	TheForceEngine/TFE_Jedi/Collision/collision.o		\
+		TheForceEngine/TFE_Jedi/IMuse/imConst.o			\
+		TheForceEngine/TFE_Jedi/IMuse/imDigitalSound.o		\
+		TheForceEngine/TFE_Jedi/IMuse/imList.o			\
+		TheForceEngine/TFE_Jedi/IMuse/imMidiCmd.o		\
+		TheForceEngine/TFE_Jedi/IMuse/imMidiPlayer.o		\
+		TheForceEngine/TFE_Jedi/IMuse/imSoundFader.o		\
+		TheForceEngine/TFE_Jedi/IMuse/imTrigger.o		\
+		TheForceEngine/TFE_Jedi/IMuse/imuse.o			\
+		TheForceEngine/TFE_Jedi/IMuse/midiData.o		\
+		TheForceEngine/TFE_Jedi/InfSystem/infState.o		\
+		TheForceEngine/TFE_Jedi/InfSystem/infSystem.o		\
+		TheForceEngine/TFE_Jedi/InfSystem/message.o		\
+		TheForceEngine/TFE_Jedi/Level/level.o			\
+		TheForceEngine/TFE_Jedi/Level/levelData.o		\
+		TheForceEngine/TFE_Jedi/Level/levelTextures.o		\
+		TheForceEngine/TFE_Jedi/Level/rfont.o			\
+		TheForceEngine/TFE_Jedi/Level/robject.o			\
+		TheForceEngine/TFE_Jedi/Level/robjData.o		\
+		TheForceEngine/TFE_Jedi/Level/roffscreenBuffer.o	\
+		TheForceEngine/TFE_Jedi/Level/rsector.o			\
+		TheForceEngine/TFE_Jedi/Level/rtexture.o		\
+		TheForceEngine/TFE_Jedi/Level/rwall.o			\
+		TheForceEngine/TFE_Jedi/Math/core_math.o		\
+		TheForceEngine/TFE_Jedi/Math/cosTable.o			\
+		TheForceEngine/TFE_Jedi/Memory/allocator.o		\
+		TheForceEngine/TFE_Jedi/Memory/list.o			\
+		TheForceEngine/TFE_Jedi/Renderer/jediRenderer.o		\
+		TheForceEngine/TFE_Jedi/Renderer/rcommon.o		\
+		TheForceEngine/TFE_Jedi/Renderer/rscanline.o		\
+		TheForceEngine/TFE_Jedi/Renderer/rsectorRender.o	\
+		TheForceEngine/TFE_Jedi/Renderer/screenDraw.o		\
+		TheForceEngine/TFE_Jedi/Renderer/virtualFramebuffer.o	\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/rclassicFixed.o			\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/rclassicFixedSharedState.o	\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/redgePairFixed.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/rflatFixed.o			\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/rlightingFixed.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/rsectorFixed.o			\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/rwallFixed.o			\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/robj3d_fixed/robj3dFixed.o			\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/robj3d_fixed/robj3dFixed_Clipping.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/robj3d_fixed/robj3dFixed_Culling.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/robj3d_fixed/robj3dFixed_PolygonDraw.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/robj3d_fixed/robj3dFixed_PolygonSetup.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Fixed/robj3d_fixed/robj3dFixed_TransformAndLighting.o	\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/rclassicFloat.o					\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/rclassicFloatSharedState.o			\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/redgePairFloat.o				\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/rflatFloat.o					\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/rlightingFloat.o				\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/rsectorFloat.o					\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/rwallFloat.o					\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/robj3d_float/robj3dFloat.o			\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/robj3d_float/robj3dFloat_Clipping.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/robj3d_float/robj3dFloat_Culling.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/robj3d_float/robj3dFloat_PolygonDraw.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/robj3d_float/robj3dFloat_PolygonSetup.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_Float/robj3d_float/robj3dFloat_TransformAndLighting.o	\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_GPU/debug.o			\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_GPU/frustum.o			\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_GPU/modelGPU.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_GPU/objectPortalPlanes.o	\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_GPU/rclassicGPU.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_GPU/renderDebug.o		\
+		TheForceEngine/TFE_Jedi/Renderer/RClassic_GPU/rsectorGPU.o		\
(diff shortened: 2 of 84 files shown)


gilmorem560:
I get the following compiler error:

```
TheForceEngine/TFE_DarkForces/darkForcesMain.cpp: In member function ‘virtual void TFE_DarkForces::DarkForces::exitGame()’:
TheForceEngine/TFE_DarkForces/darkForcesMain.cpp:377:20: error: no match for ‘operator=’ (operand types are ‘TFE_DarkForces::SharedGameState’ and ‘<brace-enclosed initializer list>’)
  377 |   s_sharedState = {};
      |                    ^
TheForceEngine/TFE_DarkForces/darkForcesMain.cpp:197:9: note: candidate: ‘constexpr TFE_DarkForces::SharedGameState& TFE_DarkForces::SharedGameState::operator=(const TFE_DarkForces::SharedGameState&)’
  197 |  struct SharedGameState
      |         ^~~~~~~~~~~~~~~
```

Not sure if there is a canonical way to clear it out, I found just putting in a second constant SharedGameState always set to the {} value then setting s_sharedState = s_nullState worked. If this was C I'd just memset(&s_sharedState, '\0', sizeof (sharedState)) but I'm pretty sure that won't navigate a pointer to a nested struct in C++ and cleanup the existing allocated memory, not really sure how that works up under the hood.

Also, was this always OpenGL 3.3? Seems my current machine (Rpi 400) won't run it :( I guess I did the work on a different computer at the time. Anywho, aside from this one fix, builds fine at least. Launches until it hits the shaders and doesn't like GLSL 1.3 on my machine. Ho hum. If I find myself on a capable piece of hardware I'll see if I can give it a spin. If nothing else I can see where all of this puts support on a mac.

mlauss:
> I get the following compiler error:
> 
> ```
> TheForceEngine/TFE_DarkForces/darkForcesMain.cpp: In member function ‘virtual void TFE_DarkForces::DarkForces::exitGame()’:
> TheForceEngine/TFE_DarkForces/darkForcesMain.cpp:377:20: error: no match for ‘operator=’ (operand types are ‘TFE_DarkForces::SharedGameState’ and ‘<brace-enclosed initializer list>’)
>   377 |   s_sharedState = {};
>       |                    ^
> TheForceEngine/TFE_DarkForces/darkForcesMain.cpp:197:9: note: candidate: ‘constexpr TFE_DarkForces::SharedGameState& TFE_DarkForces::SharedGameState::operator=(const TFE_DarkForces::SharedGameState&)’
>   197 |  struct SharedGameState
>       |         ^~~~~~~~~~~~~~~
> ```
> 
> Not sure if there is a canonical way to clear it out, I found just putting in a second constant SharedGameState always set to the {} value then setting s_sharedState = s_nullState worked. If this was C I'd just memset(&s_sharedState, '\0', sizeof (sharedState)) but I'm pretty sure that won't navigate a pointer to a nested struct in C++ and cleanup the existing allocated memory, not really sure how that works up under the hood.

Which compiler?  The code is full of initializations like these, I wonder what makes this one stand out...
I too am a fan of the memset(,0,) hammer, but yes, even compilers warn about doing that on class objects. But at least
in this case, I see no reason why the struct can't just be a plain old C struct without initializers.
Unfortunately I don't have any microsoft windows compilers to test a conversion.

Cheeseness:
FWIW, I have it compiling here with gcc 11.3.1 with no changes, but it's segfaulting in rtmidi and I don't have time to dig into that at the moment. 

> (gdb) bt
> #0  0x00007ffff7b7a270 in typeinfo for MidiInAlsa () from /lib64/librtmidi.so.4
> #1  0x0000000000446ee2 in TFE_MidiDevice::sendMessage(unsigned char, unsigned char, unsigned char) ()
> #2  0x000000000044786e in TFE_MidiPlayer::midiUpdateFunc(void*) ()
> #3  0x00007ffff749f822 in start_thread () from /lib64/libc.so.6
> #4  0x00007ffff743f450 in clone3 () from /lib64/libc.so.6

mlauss:
> 

It crashes inside your RtMidi; nothing much I can do about that.  I'm running rtmidi-5.0.0 (librtmidi.so.6), try updating if you can.

Cheeseness:
Yep, will do when I've got a chance to look into it.

JPLeBreton:
I can compile and run this branch just fine (it even loaded up an Imperial City save i'd made a few days ago in the Windows version), the only thing that doesn't work is audio (application is completely silent). These are the errors I see in the log:
```
[Startup] TFE_AudioSystem::init
RtApiJack::getDeviceInfo: error determining Jack input/output channels!
RtApiJack::getDeviceInfo: error determining Jack input/output channels!
RtApiJack::getDeviceInfo: error determining Jack input/output channels!
RtApiJack::getDeviceInfo: error determining Jack input/output channels!
[Audio] Starting up audio stream for device 'Built-in Audio Analog Stereo', id 0.
[Error : Audio] Cannot start audio stream for output device 0.
[Error : Audio] Cannot start audio stream.
```

I'm running Manjaro 64-bit (intel), latest stable branch. The version of rtmidi on my system is 5.0.0-1, from the package manager.

Given the issues RtAudio had a year ago with my PipeWire (as opposed to PulseAudio) system - https://github.com/gilmorem560/TheForceEngine/issues/2 - my first suspicion would be that. If nobody else in this thread is running their system audio via PipeWire, that might explain what I'm seeing - though I get the impression PipeWire is fairly common at this point, not just in bleeding edge distros anymore.

JPLeBreton:
I just found an AUR package that claims to fix something Jack + PipeWire related: https://aur.archlinux.org/packages/rtaudio-git

I didn't have any other programs on my system that used rtaudio, so I tried installing one https://dinisnoise.org and sure enough, it also doesn't make sound, and launching from a terminal gives the same "error determining Jack input/output channels" error. So it would appear that rtaudio just doesn't work on systems with PipeWire. So this probably isn't a Force Engine issue. I'll follow up here if I determine it is.

mlauss:
> I just found an AUR package that claims to fix something Jack + PipeWire related: https://aur.archlinux.org/packages/rtaudio-git
> 
> I didn't have any other programs on my system that used rtaudio, so I tried installing one https://dinisnoise.org and sure enough, it also doesn't make sound, and launching from a terminal gives the same "error determining Jack input/output channels" error. So it would appear that rtaudio just doesn't work on systems with PipeWire. So this probably isn't a Force Engine issue. I'll follow up here if I determine it is.

Does it work if you apply this patch:

`diff --git a/TheForceEngine/TFE_Audio/midiDevice.cpp b/TheForceEngine/TFE_Audio/midiDevice.cpp
index a824a38e..a1c33163 100644
--- a/TheForceEngine/TFE_Audio/midiDevice.cpp
+++ b/TheForceEngine/TFE_Audio/midiDevice.cpp
@@ -25,7 +25,7 @@ namespace TFE_MidiDevice
 
        bool init()
        {
-               s_midiout = new RtMidiOut();
+               s_midiout = new RtMidiOut(RtMidi::LINUX_ALSA, "TheForceEngine");
                s_midiout->setErrorCallback(midiErrorCallback);
                s_openPort = -1;
 
`

spacekomet:
Isn't it easier to just add flag -D__LINUX_ALSA__ in the makefile ? 
For example:
`CXXFLAGS=	-O2 -pipe -I./TheForceEngine -I/usr/include/SDL2 -I/usr/include/rtaudio -I/usr/include/rtmidi -D__LINUX_ALSA__`

RTAudio and RTmidi have a few of those options, which can be defined during the compiling. 
In my case when I was having problems with the sound using gilmorem560 port, I added __LINUX_PULSE__ and got it working.

JPLeBreton:
> Does it work if you apply this patch, it forces ALSA output:

Thanks! Applying those changes made sound FX work, but not MIDI.

mlauss:
> > Does it work if you apply this patch, it forces ALSA output:
> 
> Thanks! Applying those changes made sound FX work, but not MIDI.

You need a synthesizer, fluidsynth works very well for me.

JPLeBreton:
> > > Does it work if you apply this patch, it forces ALSA output:
> > 
> > Thanks! Applying those changes made sound FX work, but not MIDI.
> 
> You need a synthesizer, fluidsynth works very well for me.

I have fluidsynth installed, and I hear music from it in other applications.

Cheeseness:
> I have fluidsynth installed, and I hear music from it in other applications.

Do you have it running at the time?

eg, run this in a separate terminal:
> fluidsynth --server -a alsa -m alsa_seq -v  /usr/share/soundfonts/FluidR3_GM.sf2

JPLeBreton:
> > I have fluidsynth installed, and I hear music from it in other applications.
> 
> Do you have it running at the time?
> 
> eg, run this in a separate terminal:
> 
> > fluidsynth --server -a alsa -m alsa_seq -v  /usr/share/soundfonts/FluidR3_GM.sf2

Oh interesting, I've never had to do that before. I ran that (with a valid soundfont on my system) in another terminal and then ran TFE but still no music.

Agetian:
I can confirm that the game compiles and runs fine (using the official Dark Forces GOG Linux release files), but there is no music, regardless of whether I start Fluidsynth manually (tried several soundfonts using that command referenced above) or not. Otherwise, the game seems to play fine.

mlauss:
> > > I have fluidsynth installed, and I hear music from it in other applications.
> > 
> > 
> > Do you have it running at the time?
> > eg, run this in a separate terminal:
> > > fluidsynth --server -a alsa -m alsa_seq -v  /usr/share/soundfonts/FluidR3_GM.sf2
> 
> Oh interesting, I've never had to do that before. I ran that (with a valid soundfont on my system) in another terminal and then ran TFE but still no music.

did you see any terminal output in fluidsynth while TFE was running?   Does your running kernel have CONFIG_SND_SEQ_MIDI=y  set ?

JPLeBreton:
> did you see any terminal output in fluidsynth while TFE was running? Does your running kernel have CONFIG_SND_SEQ_MIDI=y set ?

I did not see any new terminal output in fluidsynth when TFE started up or tried to start playing music (the star wars logo). And no I do not have that env variable set. Setting it to "y" didn't seem to have any effect.

Running fluidsynth with the same command line settings + a MIDI file produced music though, so that particular fluidsynth executable is definitely capable of playing music on my system.

mlauss:
> > did you see any terminal output in fluidsynth while TFE was running? Does your running kernel have CONFIG_SND_SEQ_MIDI=y set ?
> 
> I did not see any new terminal output in fluidsynth when TFE started up or tried to start playing music (the star wars logo). And no I do not have that env variable set. Setting it to "y" didn't seem to have any effect.

It's a kernel .config entry.  Do you have the /dev/snd/seq  device node?
This device node is used by fluidsynth to listen for midi clients on, and RtMidi to connect to midi player devices.  Maybe you need to "modprobe snd_seq; modprobe snd_seq_midi" for it to appear.

> Running fluidsynth with the same command line settings + a MIDI file produced music though, so that particular fluidsynth executable is definitely capable of playing music on my system.

That's a different scenario: fluidsynth itself reads the midi file, but at least we know that fluidsynth can produce audible sound.

mlauss:
I just pushed a patch which fixes the last crash I know of.  Loading savegames while running mods should now work fine as well.

SolitudeSF:
on pipewire sound doesnt work with rtaudio 5.2.0, without any additional compiler flags, or with `__LINUX_PULSE/ALSA__`. 

`[Error : Audio] Cannot start audio stream for output device 0.`

joy of linux audio.

mlauss:
> on pipewire sound doesnt work with rtaudio 5.2.0, without any additional compiler flags, or with `__LINUX_PULSE/ALSA__`.
These are only useful for building RtAudio library itself, the windows build does that, but on linux it uses the system librtaudio.so library.

> `[Error : Audio] Cannot start audio stream for output device 0.`
Is there any line before this one?
Mine says "[Audio] Starting up audio stream for device 'TU102 High Definition Audio Controller', id 0."

> joy of linux audio.
Can't disagree here, although I haven't had any problems so far (I too run pipewire).

SolitudeSF:
> These are only useful for building RtAudio library itself, the windows build does that, but on linux it uses the system librtaudio.so library.

my bad, got a bit confused.

> Is there any line before this one?

`[Audio] Starting up audio stream for device 'Starship/Matisse HD Audio Controller Analog Stereo', id 0.`  
just normal stuff. thats my default and only output device.

mlauss:
> 

Can you apply the following patch and see what is printed in the console when audio  fails?



> > These are only useful for building RtAudio library itself, the windows build does that, but on linux it uses the system librtaudio.so library.
> 
> my bad, got a bit confused.
> 
> > Is there any line before this one?
> 
> `[Audio] Starting up audio stream for device 'Starship/Matisse HD Audio Controller Analog Stereo', id 0.` just normal stuff. thats my default and only output device. this is archlinux enviroment, if that matters.

Please apply the below patch and see whether it prints anything that explains why rtaudio is unhappy.
Thanks!

[x1.txt](https://github.com/luciusDXL/TheForceEngine/files/10332257/x1.txt)

spacekomet:
Just of  fun of it, I installed ubuntu 22.10 in virtualbox. What a nightmare...I would suggest everyone to use Manjaro :).

At the end I managed to get the sound and music working with this steps:
1. disable module snd_seq_dummy with root:
`rmmod snd_seq_dummy`
this removes dummy 14 Midi through port, which game selects by default. 
( I guess applying  patch #171 and then selecting correct midi port in the game itself should also work). 
2. run fluidsynth with pulseaudio or SDL2 drivers:
`fluidsynth --server -a sdl2 -m alsa_seq /usr/share/sounds/sf3/default-GM.sf3 ` or
`fluidsynth --server -a pulseaudio -m alsa_seq /usr/share/sounds/sf3/default-GM.sf3 ` 

At this point, checking with aconnect -l you should have one midi port 128 from fluid synth opened (without port 14) and tfelnx should play sound and music. 

Comparing to Manjaro, I have many issues with Ubuntu, like:
-  mouse doesn't work during the gameplay. It only works in TFE menus. Probably it's combination of X manager and virtualbox. When changing wayland to xorg, mouse started to work, but it was to fast to control and mirrored. 
-  if I don't run fluidsynth  beforehand,  tfelnx dumps with:
`MidiOutJack::initialize: JACK server not running?`
- at some point, I only had music in the game, but not sound....

mlauss:
>     * if I don't run fluidsynth  beforehand,  tfelnx dumps with:
>       `MidiOutJack::initialize: JACK server not running?`

I'm working on a patch to select the RtAudio/RtMidi backend in the sound settings page.
 Judging from the reports here, most distros ship both libs with JACK support enabled, which doesn't know when its not wanted.
Fortunately RtAudio/Midi allow to query the supported backends and select one at object instantiation time instead of autodetection (where jack takes precedence over alsa/pa/pw).

SolitudeSF:
> Please apply the below patch and see whether it prints anything that explains why rtaudio is unhappy.

`RtApiJack::probeDeviceOpen: the requested sample rate (11025) is different than the JACK server rate (48000).`

mlauss:
> > Please apply the below patch and see whether it prints anything that explains why rtaudio is unhappy.
> 
> `RtApiJack::probeDeviceOpen: the requested sample rate (11025) is different than the JACK server rate (48000).`

Hmm.
as root, please do the following (it deletes the "jack" api from the global pipewire config file and places the changed file in /etc/pipewire, as /usr/share/pipewire will be overridden on updates).

>mkdir -p /etc/pipewire
>cp /usr/share/pipewire/pipewire.conf /etc/pipewire
>sed -ie '/api.jack./d' /etc/pipewire/pipewire.conf

The reboot.  No guarantees that it actually fixes your problem, I know _nothing_ about pipewire.

mlauss:
> g++ (Debian 10.2.1-6) 10.2.1 20210110
```
> 
> Yeah it's pretty out of date as it goes, one of these days I'll do my updates on this system...

You should upgrade, I get the same error with gcc-10.5.  Let's just make gcc-11 the minimum.

**Comment by SolitudeSF** [comment]
>  as root, please do the following (it deletes the "jack" api from the global pipewire config file and places the changed file in /etc/pipewire, as /usr/share/pipewire will be overridden on updates).

hell naw, why would i do that.

here is how you can allow multiple sample rates for pipewire: `default.clock.allowed-rates = [ 48000, 11025 ]`. with this it works.

and as im writing this it stopped working.

can you verify that the sound keeps working if you have a sound already from some other sound client?

**Comment by mlauss** [comment]
Solitude ***@***.***> schrieb am Di., 3. Jän. 2023, 00:02:

> as root, please do the following (it deletes the "jack" api from the
> global pipewire config file and places the changed file in /etc/pipewire,
> as /usr/share/pipewire will be overridden on updates).
>
> hell naw, why would i do that.
>
Because then rtaudio will fall back to alsa/pulseaudio, which I know do
work.

here is how you can allow multiple sample rates for pipewire:
default.clock.allowed-rates
> = [ 48000 11025 ]. with this it works.
>
> and as im writing this it stopped working.
>
> can you verify that the sound keeps working if you have a sound already
> from some other sound client?
>

Yes; i havent had any problems in this regard.

The only difference between my rtaudio and yours is that I built mine
without JACK support. Most distros do not do that, and pipewire also
functions as a JACK daemon.
Unfortunately rtaudio favors jack over plain alsa/pa.
The easiest solution right now for this is to simply disable jack api
support in pipewire altogether.

.
>

**Comment by SolitudeSF** [comment]
> Because then rtaudio will fall back to alsa/pulseaudio, which I know do
work.

yes, i get it, but i will not compromise on jack support for that.

aight. you can specify the api manually. so, issue solved for me, at least.

```diff
diff --git a/TheForceEngine/TFE_Audio/audioDevice.cpp b/TheForceEngine/TFE_Audio/audioDevice.cpp
index 38d901b9..40324aba 100644
--- a/TheForceEngine/TFE_Audio/audioDevice.cpp
+++ b/TheForceEngine/TFE_Audio/audioDevice.cpp
@@ -42,7 +42,7 @@ namespace TFE_AudioDevice
 
 		try
 		{
-			s_device = new RtAudio();
+			s_device = new RtAudio(RtAudio::LINUX_PULSE);
 		}
 		catch(RtAudioError e)
 		{

```

mlauss:
> > Because then rtaudio will fall back to alsa/pulseaudio, which I know do
> > work.
> 
> yes, i get it, but i will not compromise on jack support for that.
> 
> aight. you can specify the api manually. so, issue solved for me, at least.

I've implemented Audio backend selection in the settings page.  Care to try it?
Just clone branch  "https://github.com/mlauss/TheForceEngine/tree/test-audio-select"

Agetian:
I tested the test-audio-select branch on Linux Mint 21.1 and I can confirm that I can get music to work by starting Fluidsynth and then specifying it in the MIDI Output Device option, as well as selecting Pulse as the audio backend and ALSA as the MIDI backend. Played through the first level without a problem with both sound and music.

mlauss:
Awesome, thank you for testing!

JPLeBreton:
The test-audio-select branch gives me sound and music with no patches needed - progress!

In the settings menu I can select the Pulse audio backend and hear all sound FX, and this setting is remembered across sessions.

If I select the ALSA midi backend, and run `fluidsynth --server -a pulseaudio -m alsa_seq [a soundfont of my choosing]` in a separate terminal, I hear music - although strangely most or all the instruments seem to be piano?

Cheeseness:
After compiling more recent versions of rtaudio (5.2.0) and rtmidi (5.0.0) and building with the test-audio-select branch, TFE is running and picking up my existing saves and GOG install (haven't gotten around to digging out my CD copy yet).

Aside from the issue where every instrument seems to be being played as a piano that JP notes, everything seems to be behaving happily here on Fedora.

mlauss:
Yeah I noticed the pianos as well.

I tested timidity++ as well now; using the same soundfont it not only sounds *much* richer, but fluidsynth seems to be omitting a lot of notes (or a complete track), though timidity sometimes drops some notes as well.
The "STAR-THM.MID" from DF played by timidity with aplaymidi sounds awesome (especially the final parts which are played while the credits roll), with the correct instruments (horns, mainly). However when TFE plays it, it sounds like its played by a muted piano.  Maybe there's something missing wrt. patch changes, but I don't know any MIDI.

tl;dr: use timidity++ instead of fluidsynth for more accurate music.

Cheeseness:
> I tested timidity++ as well now; using the same soundfont it not only sounds _much_ richer

Interesting. I hear only pianos when running the game's midi through timidity. What version are you running, and are you explicitly specifying a soundfont? (version 2.15.0 here)

mlauss:
> > I tested timidity++ as well now; using the same soundfont it not only sounds _much_ richer
> 
> Interesting. I hear mostly pianos (plus some percussion - I think the same as what I was getting with fluidsynth) when running TFE's midi through timidity. What version are you running, and are you explicitly specifying a soundfont? (version 2.15.0 here)

Yeah my bad, it's still just pianos.   There's something bad happening after midi commands are handed off to RtMidi. As far as I can tell, TFE is doing the right thing.

mlauss:
This little patch seems to fix it for me.  It seems RtMidi and/or ALSA get confused by the 2nd byte for
the Channel Control and Program Change messages.

diff --git a/TheForceEngine/TFE_Audio/midiDevice.cpp b/TheForceEngine/TFE_Audio/midiDevice.cpp
index 91715927..21e912cd 100644
--- a/TheForceEngine/TFE_Audio/midiDevice.cpp
+++ b/TheForceEngine/TFE_Audio/midiDevice.cpp
@@ -137,6 +137,9 @@ namespace TFE_MidiDevice
 
        void sendMessage(const u8* msg, u32 size)
        {
+               u8 x = msg[0] & 0xf0;
+               if ((x == 0xc0) || (x == 0xd0))
+                       size = 2;
                if (s_openPort > 0) { s_midiout->sendMessage(msg, (size_t)size); }
        }

Cheeseness:
Weird! I don't have any useful insights, but that makes instrumentation behave for me as well.

Agetian:
Tested the latest changes with Timidity++ in daemon mode (-iA) on Linux Mint 21.1, it works really fantastic compared to how things were before, I have rich instrumental music now with multiple instruments. Sounds much nicer than it was before! :+1:

mlauss:
> Edit: If this is an ALSA-specific problem, it might be worth putting in some kind of guard or leaving a comment for future developers in case environments change?

Technically TFE behaviour was a violation of the midi protocol, but since the additional byte was a zero, and I cannot find any command for a zero byte in the public midi docs, I bet the windows midi parser just throws it away, and it never was an issue.  The alsa parser seems is not that forgiving.

SolitudeSF:
could  you make TFE use proper directory for its data, like XDG_DATA_HOME, instead of home?

gilmorem560:
> could you make TFE use proper directory for its data, like XDG_DATA_HOME, instead of home?

I would actually advocate *not* doing this. $HOME is in the POSIX standard, $XDG_DATA_HOME is not, so this would break support for UNIXes that don't implement XDG, of which there are many.

SolitudeSF:
no one expects that every machine is going to have XDG_DATA_HOME set, every single program that ever used it fallsback to ~/.local/share. and  using `HOME` to store your data without even making it hidden is beyond comprehension.

gilmorem560:
I do it on my machine and it doesn't burst into flames, burn my bank account down, and render a plague unto my house...

SolitudeSF:
im happy for you. every program is trying to patch in support for proper data storage, but you know better.

gilmorem560:
"Proper data separation" is chiefly a decision to be made by the person owning the data store. I don't want an application over-rulling me saying where my data needs to go based on a subset behavior like XDG instead of POSIX which uses the global variable HOME that has existed since the early 70s if not before. Not every machine is a GNU/Linux running Potteringware catering to a lack of understanding of how the UNIX ecosystem actually works.

SolitudeSF:
crazy, havent seen such bold ignorance be paraded as indepth expertise. you're the one the rare few people i've seen that for some reason are viciously opposed to XDG as if it appealing to its standard makes your system compromised in some way. every other program you can find on github that stores some data uses XDG, and other half defaults to `.config`/`.local/share`. only the most old stubborn legacy piece of shit still resorts to putting stuff straight in `HOME`. i dont know why i need to explain this. this is the stupidest shit i've seen all week.

Cheeseness:
> every single program that ever used it fallsback to ~/.local/share

This behaviour is part of the [XDG Base Dir spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Anything that doesn't have a fallback that explicitly uses the $HOME variable isn't adhering to the standard.

>  $XDG_DATA_HOME defines the base directory relative to which user-specific data files should be stored. If $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used.

 &nbsp; 

> what you want then is a $TFE_DATA_DIR environment variable that you get to set to $XDG_DATA_HOME, I get to set to $HOME, etc.

I think this is a reasonable solution. Having it default to `$XDG_DATA_HOME/TheForceEngine` when not set with a fallback to `$HOME/.local/share/TheForceEngine` and just `$HOME/TheForceEngine` if ~/.local/share doesn't exist feels like it gives the best of both worlds while giving sensible defaults and respecting the XDG Base Dir standard **and** giving users a good degree of flexibility.

 &nbsp; 

Of course, it would be nice to separate user data from config data as well (which the XDG Base Dir spec also addresses), but one hurdle at a time.

gilmorem560:
> crazy, havent seen such bold ignorance be paraded as indepth expertise. you're the one the rare few people i've seen that for some reason are viciously opposed to XDG as if it appealing to its standard makes your system compromised in some way. every other program you can find on github that stores some data uses XDG, and other half defaults to `.config`/`.local/share`. only the most old stubborn legacy piece of shit still resorts to putting stuff straight in `HOME`. i dont know why i need to explain this. this is the stupidest shit i've seen all week.

Then make those decisions in your codebases or make a patch to the changes I made. You're allowed to do such things without castigating another person for adhering to much more globally vetted standards than XDG, especially when that person did work you're not benefitting from to complain about in the first place. Again, some of us use systems other than GNU/Pottering and accept that POSIX is still the lowest common denominator.

Agetian:
As a possible constructive idea regarding the storage of configuration, maybe it's possible to make this configurable, so that everyone can be happy, no matter if they prefer storage in the Home directory directly or in one of the common Linux configuration locations such as .local/share or .config? Just an idea though. Personally I don't mind either way, having TFE on Linux is worth it for me either way, hehe - I can certainly live with an extra subdirectory in Home if that's what it takes to play the game :+1:

Agetian:
I would like to report a slight issue with setting the game path on Linux, I'm not sure if this is specific to my system (which is Linux Mint 21.1) or to the case-sensitive ext4 file system in general. When specifying the initial game path in the configuration, The Force Engine asks to locate the file called DARK.GOB. However, when I browse into the game folder (I'm using the official Linux version of the GOG release of Dark Forces), I don't see the file (DARK.GOB) and I can't select it. I suspect that it's due to the fact that the engine is actually looking (at this point) for dark.gob, lowercase - and indeed, if I temporarily rename the file to lowercase (or create a dummy lowercase dark.gob alongside the properly capitalized DARK.GOB), the game will allow me to select the file and thus the folder. I'll need to remove that dummy file (or rename the gob file back to DARK.GOB). Maybe the issue is that the file extension is expected to be lowercase at that point somewhere in the code? Not sure what exactly is up, but it's definitely slightly inconsistent in what it expects at that point and what the game itself expects (the capitalized file name).

mlauss:
I think this depends on the desktop environment, as TFE shows (at least on KDE) the native file dialog, and at least I can locate all the files with uppercase extensions just fine.  Do you use gnome?

Does the below patch fix your file dialog?

```diff
diff --git a/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp b/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
index b4d1fcb7..024ba20b 100644
--- a/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
+++ b/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
@@ -1010,8 +1010,8 @@ namespace TFE_FrontEndUI
                        };
                        const std::vector<std::string> filters[]=
                        {
-                               { "GOB Archive", "*.gob", "Executable", "*.exe" },
-                               { "LAB Archive", "*.lab", "Executable", "*.exe" },
+                               { "GOB Archive", "*.GOB", "Executable", "*.EXE" },
+                               { "LAB Archive", "*.LAB", "Executable", "*.EXE" },
                        };

```

HandMeBacon:
> crazy, havent seen such bold ignorance be paraded as indepth expertise. you're the one the rare few people i've seen that for some reason are viciously opposed to XDG as if appealing to its standard makes your system compromised in some way. every other program you can find on github that stores some data uses XDG, and other half defaults to `.config`/`.local/share`. only the most old stubborn legacy piece of shit still resorts to putting stuff straight in `HOME`. i dont know why i need to explain this. this is the stupidest shit i've seen all week.

Even if you were 100% right, the way you've been speaking is rude and unhelpful. Please do everyone a favor and learn to have conversations like an adult before fielding your suggestions.

gilmorem560:
My rec from working with file pickers before, I'd just err on the side of caution and have catches for upper and lower-case if you can't do a case-insensitive compare. If that's not feasible for some reason, then ensuring the picker will always have a fallback *.* "All Files" option available as well might suffice.

SeongGino:
For what it's worth, fullscreening doesn't seem to set the relevant flag to the WM indicating it's a fullscreen window, and instead it's just a borderless window spanning the screen's size but gets covered by panels/docks (at least on Plasma, 5.26.5).

I would also vouch for implementing some checks to allow storage in `~/.local/share` or `~/.config` or equivalent (or at the very least make the bloody folder a dotfile) but I believe enough's been said on that subject already at this stage.

mlauss:
> For what it's worth, fullscreening doesn't seem to set the relevant flag to the WM indicating it's a fullscreen window, and instead it's just a borderless window spanning the screen's size but gets covered by panels/docks (at least on Plasma, 5.26.5).

Let me look into that...

> I would also vouch for implementing some checks to allow storage in `~/.local/share` or `~/.config` or equivalent (or at the very least make the bloody folder a dotfile) but I believe enough's been said on that subject already at this stage.

So I don't have any strong opinion on that matter; for SteamDeck compatibility (which I think luciusDXL is aiming at) the 
$XDG_DATA_HOME approach is probably the best solution.  I personally prefer Matthew Gilmores initial implementation,
simply because I have to type less to access the TFE directory :)    The port currently does implement "~/.local/share/" which
is as far as I can see, unused though.  Maybe it's a remnant of the Windows-way of separating user folders.

xcom169:
We need fluidsyth package installed separately on the Linux OS or it will be compiled as a static Dll?

mlauss:
> We need fluidsyth package installed separately on the Linux OS or it will be compiled as a static Dll?

At the moment you need either fluidsynth or timidity++ installed separately and running to hear music.
It's not linked in or anything, rather it acts like a hardware synthesizer (remember the MT-32? good times...)

Digital Audio is currently operating at 11kHz because that's the samplerate of the DF sound effects. Add some
internal audio mixing engine, and I see no reason why fluidsynth (or a better-sounding synth) cannot be
added as a built-in midi synth.

mlauss:
> > could you make TFE use proper directory for its data, like XDG_DATA_HOME, instead of home?
> 
> I would actually advocate _not_ doing this. $HOME is in the POSIX standard, $XDG_DATA_HOME is not, so this would break support for UNIXes that don't implement XDG, of which there are many.
> 
> Edit: A compromise may be to coalesce, check for $XDG_DATA_HOME and fall back to $HOME if it isn't there. Either way, $HOME was a very intentional choice, not an oversight.

I made a "linux3" branch, where I took a slightly different approach:
- check for Envvar "TFE_DATA_HOME".  If it's set, use this as a prefix path for the final "TheForceEngine" dir,
- otherwise check for the XDG_DATA_HOME dir, use that one; and if it's unset,
- just use .local/share/ as prefix.
On most modern linux distros, I think the XDG_DATA_HOME is the right approach.

So for me and @gilmorem560  we need to "export TFE_DATA_HOME=/" or symlink to .local/..

Hm?

mlauss:
> For what it's worth, fullscreening doesn't seem to set the relevant flag to the WM indicating it's a fullscreen window, and instead it's just a borderless window spanning the screen's size but gets covered by panels/docks (at least on Plasma, 5.26.5).
> 

Yup, this 2-liner fixes it for me:

```diff

diff --git a/TheForceEngine/TFE_RenderBackend/Win32OpenGL/renderBackend.cpp b/TheForceEngine/TFE_RenderBackend/Win32OpenGL/renderBackend.cpp
index 5d2c2861..3fa58f23 100644
--- a/TheForceEngine/TFE_RenderBackend/Win32OpenGL/renderBackend.cpp
+++ b/TheForceEngine/TFE_RenderBackend/Win32OpenGL/renderBackend.cpp
@@ -374,6 +374,7 @@ namespace TFE_RenderBackend
 
                        SDL_SetWindowSize((SDL_Window*)m_window, m_windowState.monitorWidth, m_windowState.monitorHeight);
                        SDL_SetWindowPosition((SDL_Window*)m_window, s_displayBounds[displayIndex].x, s_displayBounds[displayIndex].y);
+                       SDL_SetWindowFullscreen((SDL_Window*)m_window, SDL_WINDOW_FULLSCREEN);
 
                        m_windowState.width  = m_windowState.monitorWidth;
                        m_windowState.height = m_windowState.monitorHeight;
@@ -382,6 +383,7 @@ namespace TFE_RenderBackend
                {
                        m_windowState.flags &= ~WINFLAG_FULLSCREEN;
 
+                       SDL_SetWindowFullscreen((SDL_Window*)m_window, 0);
                        SDL_RestoreWindow((SDL_Window*)m_window);
                        SDL_SetWindowResizable((SDL_Window*)m_window, SDL_TRUE);
                        SDL_SetWindowBordered((SDL_Window*)m_window, SDL_TRUE);

```

QmwJlHuSg9pa:
> > We need fluidsyth package installed separately on the Linux OS or it will be compiled as a static Dll?
> 
> At the moment you need either fluidsynth or timidity++ installed separately and running to hear music. It's not linked in or anything, rather it acts like a hardware synthesizer (remember the MT-32? good times...)
> 
> Digital Audio is currently operating at 11kHz because that's the samplerate of the DF sound effects. Add some internal audio mixing engine, and I see no reason why fluidsynth (or a better-sounding synth) cannot be added as a built-in midi synth. EDIT: the above can be achieved by using sdl2_audio and sdl2_mixer instead of Rtaudio. If there's any interest, I'll take a stab at implementing it.

Had you considered OpenAL Soft, instead? This is a likely route to take for #159, at least.

Ditto for mt32emu, if Roland MT-32/CM-32L accuracy is desirable? Though, most users could probably use Munt externally, and ROMs would need to be provided.

mlauss:
> > > We need fluidsyth package installed separately on the Linux OS or it will be compiled as a static Dll?
> > 
> > 
> > At the moment you need either fluidsynth or timidity++ installed separately and running to hear music. It's not linked in or anything, rather it acts like a hardware synthesizer (remember the MT-32? good times...)
> > Digital Audio is currently operating at 11kHz because that's the samplerate of the DF sound effects. Add some internal audio mixing engine, and I see no reason why fluidsynth (or a better-sounding synth) cannot be added as a built-in midi synth. EDIT: the above can be achieved by using sdl2_audio and sdl2_mixer instead of Rtaudio. If there's any interest, I'll take a stab at implementing it.
> 
> Had you considered OpenAL Soft, instead? This is a likely route to take for #159, at least.

nope, but i'll take a look

> Ditto for mt32emu, if Roland MT-32/CM-32L accuracy is desirable? Though, most users could probably use Munt externally, and ROMs would need to be provided.

You can already use mt32emu_alsa, just like fluidsynth or timidity.

mlauss:
Please refer to #188

PR #177, opened 2022-12-31
WIP (DO NOT MERGE): Switch away from relative mouse mode in menus

beholdnec:
In this branch, The Force Engine will switch away from relative mouse mode when opening the agent menu, the escape menu and the PDA. This makes the cursor acceleration and sensitivity behave identically to the operating system when a menu is active.

This branch is not suitable for merging. It has the following known issues:
- The OS cursor becomes visible in addition to the game cursor.
- Cursor might be positioned wrong in some screen configurations, such as widescreen.

I'll need guidance to solve these issues.


beholdnec:
This should be ready for testing now.

PR #205, opened 2023-01-27
RClassic_GPU: modelGPU refinements

### LINKED ISSUE #203: [Bug] Linux port, crash in Dark Tide 2
Hello

I'm experiencing crash in mod Dark Tide 2, after I pick the scuba dive and swim in the water. 

Mod is downloaded from DF-21, https://df-21.net/downloads/levels/dtide2/dtide2_modern.zip

Here is compressed save. Unzip it and put it in the save folder. 
[save007.tfe.zip](https://github.com/luciusDXL/TheForceEngine/files/10475604/save007.tfe.zip)
After it's loaded, turn right and swim up to the furthest wall, and then turn around. 

Here is the log (however the error was not saved):
[the_force_engine_log.txt](https://github.com/luciusDXL/TheForceEngine/files/10475610/the_force_engine_log.txt)

Here is the actual error shown in the console:
```
tfelnx: /home/user/ForceEngine/TheForceEngine-audio/TheForceEngine/TheForceEngine/TFE_Jedi/Renderer/RClassic_GPU/modelGPU.cpp:687: void TFE_Jedi::model_add(void*, JediModel*, Vec3f, fixed16_16*, f32, Vec2f, Vec2f, u32): Assertion `0' failed.
zsh: IOT instruction (core dumped)  ./tfelnx
```

Using linux port https://github.com/mlauss2/TheForceEngine/tree/linux3-audio-next

Anyone else with the same problem?

mlauss2:
This set of changes  does the following:
- remove the hardcoded object limit from the modelGPU rendering system (fixes #203), replaces it with dynamic allocation and freeing of objects.

Additionally, since I was staring at this code:
- static-ize all internally-used functions, which, at least with gcc, results in slightly smaller code.
- get rid of all but one of the containers for the model lists: replace the indices used with pointers to the objects themselves. 
- pack the state of the model building into a struct which is then passed around the user functions.

EDIT: reworded since more changes were added after the original commit.

### DIFF
--- a/TheForceEngine/TFE_Asset/modelAsset_jedi.cpp
+++ b/TheForceEngine/TFE_Asset/modelAsset_jedi.cpp
@@ -269,6 +269,10 @@ namespace TFE_Model_Jedi
 	{
 		// Memory will get freed with the memory region automatically.
 		s_models[pool].clear();
+		
+		// free the memory of each models' drawId object
+		std::for_each(s_modelList[pool].begin(), s_modelList[pool].end(),
+				[](JediModel *m){ if (m->drawId) free(m->drawId); });
 		s_modelList[pool].clear();
 		s_modelNames[pool].clear();
 	}
--- a/TheForceEngine/TFE_Asset/modelAsset_jedi.cpp
+++ b/TheForceEngine/TFE_Asset/modelAsset_jedi.cpp
@@ -348,7 +348,7 @@ namespace TFE_Model_Jedi
 		model->textureCount = 0;
 		model->textures = nullptr;
 		model->radius = 0;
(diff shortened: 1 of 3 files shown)


mlauss2:
Oh and there's also no perf impact: at 3440x1440, I get 5000fps with or without this.

luciusDXL (approved):
Seems fine, though I might tweak some aspects afterward.

mlauss2:
> Seems fine, though I might tweak some aspects afterward.

Oh thanks!  Yeah, I'll try to adapt to your codingstyle, but it's hard to change one's habits..

PR #206, opened 2023-01-29
Linux: update CMake

mlauss2:
- remove the "VERSION" tag, it's unused and always out of date anyway.
- Linux: raise minimum version requirements for RtAudio and RtMidi to the versions actually used during testing of the port.  No advantage in supporting outdated and untested versions.


luciusDXL (approved):
Thanks.

PR #214, opened 2023-02-02
Added .desktop file and icon

RC2225:
Added for Linux .desktop file. The 256x256 icon is from the Windows .ico.
I don't have much experience with CMake, so I don't know if the install part is best practices. I doubt it, but it works.


CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ if(WIN32)
         set(CMAKE_INSTALL_LIBDIR ".")
         set(CMAKE_INSTALL_DATADIR ".")
 else()
+	set(TFE_ICONDIR "/usr/local/share/icons/hicolor")
```
mlauss2, line 33: Don't set absolute paths in here, this will break if you do a "make DESTDIR=$PWD/image install".
Instead, above line 37, do something like "set(TFE_ICONDIR "share/icons/hicolor")", same for desktopdir.

TheForceEngine/TheForceEngine.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Name=The Force Engine
+Comment=Reimplementation of the Jedi Engine for Star Wars: Dark Forces and Outlaws
+Comment[de]=Reimplementation der Jedi Engine benutzt von Star Wars: Dark Forces und Outlaws
+Exec=tfelnx
+Type=Application
+Categories=Game;Shooter
+Path=/usr/local/bin
```
mlauss2, line 8: Again, this will break when another install prefix is specified.  Use cmakes install prefix and let this part be generated by cmake itself.

mlauss2 (commented):
I like it, but
a) don't use absolute paths as it breaks e.g. make "DESTDIR=$PWD/image install"
b) use the CMake install prefix, as someone might want to install it in /usr/share (like me),

RC2225:
Is there a variable to get the bin name tfelnx into the .desktop file? Or should I replace `set_target_properties(tfe PROPERTIES OUTPUT_NAME "tfelinux")` with  `set_target_properties(tfe PROPERTIES OUTPUT_NAME ${TFE_BINNAME})` and define the variable where the other Linux specific vars are defined?

mlauss2:
I read through the .desktop file description, I think you can drop the "Path" line altogether (it defines the initial cwd, which tfe resets anyway), and just use the binary name for Exec, i.e "Exec=tfelnx", which should work for most distros.

mlauss2:
If you do just this:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 4169cf01..277a93a5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,3 +103,9 @@ install(DIRECTORY
           GROUP_READ GROUP_EXECUTE GROUP_WRITE
           WORLD_READ WORLD_EXECUTE
 )
+
+# Linux .desktop files
+if(LINUX)
+       install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/TheForceEngine/TheForceEngine.desktop" DESTINATION "share/applications")
+       install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/TheForceEngine/TheForceEngine.png" DESTINATION "share/icons/hicolor/256x256/apps")
+endif()
```

And add a static .desktop file:
```diff

diff --git a/TheForceEngine/TheForceEngine.desktop b/TheForceEngine/TheForceEngine.desktop
new file mode 100644
index 00000000..63191111
--- /dev/null
+++ b/TheForceEngine/TheForceEngine.desktop
@@ -0,0 +1,8 @@
+[Desktop Entry]
+Type=Application
+Name=The Force Engine
+Comment=Modern Jedi Engine Replacement for LucasArts' DarkForces and Outlaws.
+Categories=Game;ActionGame;Shooter;
+Exec=tfelnx
+Icon=TheForceEngine
+StartupWMClass=tfelnx
```

Then it works well enough.  I can package it on Gentoo, it works if installed in /usr/bin and /usr/local/bin, and a desktop entry shows up in KDE automatically as well.
Should be good enough for almost every mainstream distro.

RC2225:
Added the ActionGame Category. Tested with rofi and Gnome.

CMakeLists.txt
```diff
@@ -65,7 +66,6 @@ elseif(LINUX)
 				${RTAUDIO_LIBRARIES}
 				${RTMIDI_LIBRARIES}
 	)
-
```
mlauss2, line 68: drop that please

mlauss2 (approved):
Other than the stray empty line removal, I like it.

PR #217, opened 2023-02-03
DF: implement Keyboard keys for other languages

### LINKED ISSUE #212: Only English keyboard shortcuts are used
Different languages use different highlighted characters to navigate the menu. As an example in German Ja (yes) it would be J and not Y.

mlauss2:
Other language versions of DF use different keys in the Briefing, Agent, and Escape Menu screens.

Parse the "HOTKEYS.MSG" file which is different for each language version to get to know the right
hotkeys to use.

Fixes #212


luciusDXL (approved):
Thanks for fixing that. :)

mlauss2:
> Thanks for fixing that. :)

You know, working on your engine is the most programming fun I've had in quite some time.  And I get to play Dark Forces for debugging :)  So thanks for that!

PR #222, opened 2023-02-04
fix MIDI device list and selection

### LINKED ISSUE #200: [Feature Request] better midi device detection
Of course, the addition of Midi devices greatly improves the sound quality of Dark Forces, however, TFE, for some reason, only detects one device (usually, what the operating system detects by default), but what if the user has more on one?; There is a wide variety of Soundfont available on the internet and programs that allow the creation of Midi devices that can be associated with these Soundfont, Virtual midi synth and Fluidsynth are good examples of this. The implementation of detection of other Midi devices (in addition to the standard one), would allow a greater variety of options, without having to leave TFE, as already happens with GZDoom, for example

tfe_screenshot_Fri-Jan-20-16_10_13-2023_0
 here is a current screenshot of the Midi device selection in TFE

tfe_screenshot_Wed-Jan-18-16_15_19-2023_0
here is a screenshot of a version of TFE created by Sythmartir in PR, to add this feature.

I know that there is a technical challenge, no doubt, but if that implementation is not suitable for the project, create a more suitable one, thus bringing more sound options to TFE.

mlauss2:
- the midi device list only showed the first entry, fix that.
- upon selecting a different midi port, close the open connection first. Gets rid of an error I see on Linux.

Should fix #200 and make #171 irrelevant.

### DIFF
--- a/TheForceEngine/TFE_Audio/midiDevice.cpp
+++ b/TheForceEngine/TFE_Audio/midiDevice.cpp
@@ -71,6 +71,7 @@ namespace TFE_MidiDevice
 		}
 		if (index != s_openPort && index >= 0 && index < getDeviceCount())
 		{
+			s_midiout->closePort();
 			s_midiout->openPort(index);
 			s_openPort = (s32)index;
 			return true;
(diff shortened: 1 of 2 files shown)


luciusDXL (approved):
Thanks, I have been meaning to get to this.

BielBdeLuna:
in Linux midi still doesn't play: #226

PR #223, opened 2023-02-05
DarkForces: optionally play GEARUP.LFD animation

mlauss2:
Optionally play the GEARUP.LFD animation before mission start, if it is available.

Signed-off-by: Manuel Lauss <manuel.lauss@gmail.com>


luciusDXL:
I am planning on adding an "autoload" folder to handle mods that should always be loaded (such as this and the community patch) - so I'm going to hold off on this until that is implemented.

PR #225, opened 2023-02-05
DarkForces: apply correct palette to escape menu.

### LINKED ISSUE #211: Quit to DOS confirmation dialogs in German version has different colors
When clicking Programm beenden (Quit to DOS) the text is rendered pink with some red "ghosting".
I switched for fun to the Spanish version, where it happens partially in the Abort mission confirmation dialog. 
Tested on Linux and Windows.
Pink text

mlauss2:
Apply the menu.pltt palette to the escape menu instead of the secbase level palette.
This fixes the slightly magenta text in the "Quit to Dos are you sure" messagebox of the german version of the game.

Fixes #211

### DIFF
--- a/TheForceEngine/TFE_DarkForces/GameUI/escapeMenu.cpp
+++ b/TheForceEngine/TFE_DarkForces/GameUI/escapeMenu.cpp
@@ -135,7 +135,7 @@ namespace TFE_DarkForces
 		return range;
 	}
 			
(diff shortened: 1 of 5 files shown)


luciusDXL:
I applied the fix myself so I could make the escape menu palette internal and avoid needing to reference it outside of the escape menu. Thanks for the PR though, the fix was obvious in hindsight - I'm not sure what I was thinking using the Secbase palette.

PR #233, opened 2023-02-07
Linux: misc fixes

### LINKED ISSUE #229: [Linux] Dark Tide 4 / M1: multiple bugs
Dark Tide 4, mission 1 (dtide4_modern2.zip) right after the first room with the ceiling fan on to the next room, the floor is transparent, makes it hard to jump over the actual hole.

image

### LINKED ISSUE #231: Linux compile instructions
Hi, since the release download is only made for Windows, can we get some proper compile / build instructions added if there's not going to be a pre-made Linux build? :+1: 

The new release says:

> Updated the README with Linux build and run instructions.

But it doesn't? It mentions the needed libs but no instructions on how to actually build and run it

mlauss2:
- Add step-by-step instructions on how to build and instal TFE Linux from source. Fixes #231 
- change directory/file enumeration to not treat symlinks as special: Fixes #229 
- make all file/dir existence checks case-agnostic: Since DF and Outlaws are DOS/Windows applications, they are used to case-agnostic filesystems, so the linux implementation must do the extra steps to be so as well (noticed because I could not select the "Outlaws.lab" file).

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,17 @@ This can be overridden by defining the "__TFE_DATA_HOME__" environment variable.
 ##### Building from Source
 * [CMake](https://cmake.org) 3.12 or higher to build the source.
 * GCC-11 and newer or equivalent clang version.
+* How to build
+	* unpack the source or fetch from github
+	* create a build directory and chdir into it:
+__mkdir tfe-build; cd tfe-build__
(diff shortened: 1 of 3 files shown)


luciusDXL (approved):
Approved.

PR #235, opened 2023-02-08
Linux/CMake: also install the Mods folder from source

mlauss2:
It contains the adjustable HUD Mod, which was not installed by default and caused crashes before lucius fixed the code.


luciusDXL (approved):
Thanks.

PR #237, opened 2023-02-09
Linux: name executable "TheForceEngine"

mlauss2:
Call the final executable on Linux "TheForceEngine", just like on Windows, for consistency on all supported platforms.

While "tfelnx" is short, it does not immediately remind one of what it is about.


mlauss2:
Oh, we can't

mlauss2:
Fixed.  Can't have caps the output filename because the executable and a build subdir would cause a name clash.

PR #241, opened 2023-02-10
Cleats and swimming sounds

jerethk:
Adds code for ice-cleat and swimming sound effects, which were included with the game but unused in the original.
The feature is a toggleable game setting, like autorun.

Note: the 4 cleats VOC files need to be shipped separately. The original game has a single VOC file with all 4 combined; I have split them so the interval can be altered to match how fast the player is moving.

### DIFF
--- a/TheForceEngine/TFE_DarkForces/player.cpp
+++ b/TheForceEngine/TFE_DarkForces/player.cpp
@@ -1379,7 +1379,7 @@ namespace TFE_DarkForces
 			}
 			else
 			{
-				if (!s_playerInWater)
+				if (!s_playerInWater || !(TFE_Settings::getGameSettings()->df_movementSounds))
 				{
 					sound_play(s_jumpSoundSource);
 				}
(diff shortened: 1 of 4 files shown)


luciusDXL:
I see the PR and looked over the code. It seems fine - but I have to think about how to handle non-vanilla changes like this before accepting. The change isn't that big in this case, but changes like this can add up and make it harder to refactor the code so that it is properly moddable. So give me a few days to figure it out.

FitzRoyX:
I'm all for fixes in the main repo, but the problem with restoring unused content is that it may not be a fix. They could've found the effects too annoying and left them out intentionally. They are, after all, walking sounds and by nature repeating.

jerethk:
> I'm all for fixes in the main repo, but the problem with restoring unused content is that it may not be a fix. They could've found the effects too annoying and left them out intentionally. They are, after all, walking sounds and by nature repeating.

I put it in as an option which can be turned on or off, like the other enhancements that have been added so far (eg. display secret message and autorun). We don't _really_ know why the effects were cut. One thing we do know for sure is that they were racing the clock and rushed leading up to release.

The subsequent generation of games, like Jedi Knight, had footsteps on every surface but that wasn't considered annoying. My patch only adds them for the ice cleats, which are only used in 1 level out of 14. I would be much more hesitant to add universal footsteps.

PR #242, opened 2023-02-12
Add comments categorising keywords and noting those which are not implemented

jerethk:
I started out doing this to identify the ones which are un-implemented
But ended up labelling all the INF and logic ones too
Let me know if you think it's helpful. If you'd prefer I can wind it back to just identifying the unimplemented ones


luciusDXL (approved):
Seems fine.

PR #255, opened 2023-02-21
CMake: Linux: don't use ln absolute path

### LINKED ISSUE #254: Regarding linux and installation
Hi!
Is there some way to run TheForceEngine on linux without issuing "make install"?
As it seems for now the files must reside in /usr/local/bin, /usr/local/share/TheForceEngine/ or you can add another hardcoded path via parameters before building.
If think it would be neat to have the possibility to run TheForceEngine from one folder without having a hardcoded path.
Makes portability easier.

Thanks in advance

mlauss2:
Just assume ln is somewhere in the user's PATH.

Fixes #254


luciusDXL (approved):

moether:
Consider to use the more generic cmake equivalent:
 https://cmake.org/cmake/help/latest/command/file.html#create-link

moether:
Oops. I see you are bound to CMake 3.12:
cmake_minimum_required(VERSION 3.12 FATAL_ERROR)

The createlink file command was added with 3.14.

So just ignore my previous comment.

mlauss2:
> Oops. I see you are bound to CMake 3.12: cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
> 
> The createlink file command was added with 3.14.
> 
> So just ignore my previous comment.

Care to create a patch?  Just raise the minimum version to 3.14 with it.

moether:
Here you go: https://github.com/moether/TheForceEngine/commit/98b3a3d8d22c1b714b991205ebfbbc5cefea132e
.gitignore was just updated to suit my personal needs as am using vscode + according cmake extension. Perhaps others do as well, so I decided to contribute this change too

PR #273, opened 2023-06-07
IMuse: clamp priority parameter to 127

mlauss2:
Sometimes the priority computed is larger than the allowed 127 by MiDi, and this confuses the Linux ALSA MIDI Parser.  Ensure it's never larger than 127.
Fixes a ton of "incomplete message!" MIDI errors in the log when System Midi is in use on Linux.


luciusDXL (approved):
Looks good, thanks.

PR #277, opened 2023-06-18
Fix a bug in getDisplayIndex locating the current display

tkuester:
Seems there was a typo, where `y` was compared to `s_displayBounds[i].x`!


mlauss2 (approved):
yeah seems like a copy-paste mistake. Although X is usually larger than Y and therefore it never gave any ill effects.

luciusDXL:
I fixed this already, so I'm going to close. Thanks for pointing it out though. :)

PR #279, opened 2023-06-20
Linux: improve crash handler

mlauss2:
Slightly improve the linux crash handler:
- print signal information
- print a backtrace if possible
- avoid own signal handling when being run under debuggers/tracers/...


luciusDXL (approved):
Thanks for improving the crash handler. :)

PR #282, opened 2023-07-02
Big endian fixes

BSzili:
I started working on porting TFE to big endian systems. I focused mostly on loading the original game data, editor-only features were not tested. This should allow TFE to run on various alternate systems, like PS3, XB360, Wii, WiiU, Amiga and its derivatives.
To avoid reinventing the wheel I wrapped SDL_endian.h in a header (TFE_System/endian.h), so it can be replaced with custom implementations if necessary.

### DIFF
--- a/TheForceEngine/TFE_DarkForces/agent.cpp
+++ b/TheForceEngine/TFE_DarkForces/agent.cpp
@@ -451,11 +451,11 @@ namespace TFE_DarkForces
 		{
 			return JFALSE;
 		}
-		saveData->agentData.selectedMission = TFE_Endian::le32_to_cpu(saveData->agentData.selectedMission);
-		saveData->agentData.nextMission = TFE_Endian::le32_to_cpu(saveData->agentData.nextMission);
+		saveData->agentData.selectedMission = TFE_Endian::cpu_to_le32(saveData->agentData.selectedMission);
+		saveData->agentData.nextMission = TFE_Endian::cpu_to_le32(saveData->agentData.nextMission);
 		for (s32 i = 0; i < 140; i++)
 		{
-			saveData->ammo[i] = TFE_Endian::le32_to_cpu(saveData->ammo[i]);
+			saveData->ammo[i] = TFE_Endian::cpu_to_le32(saveData->ammo[i]);
 		}
 		file->writeBuffer(saveData, sizeof(LevelSaveData));
 		return JTRUE;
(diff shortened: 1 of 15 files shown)


mlauss2:
Great idea! What do you think about calling the functions "readLE32()" or similar (or le32_to_cpu() like Linux kernel uses), to indicate to the casual code reader that the attempt is to read 32-bit little-endian formatted data into host cpu endianness?

BSzili:
The name readLE32 implies some kind of I/O for me, also these functions are used for saving too, e.g. for DARKPILO.CFG. I didn't put much tought into the naming convention though, just copied what they are called in SDL2, so le32_to_cpu and co are fine by me. I'll look up how they are named in the linux kernel and replace them here.

BSzili:
I renamed the functions and rebased the PR. The float variants had no Linux kernel equivalents, so I took some liberties there. Let me know what you think.

TheForceEngine/TFE_DarkForces/agent.cpp
```diff
@@ -434,16 +435,28 @@ namespace TFE_DarkForces
...
 		if (!file->seek(fileOffset))
 		{
 			return JFALSE;
 		}
+		saveData->agentData.selectedMission = TFE_Endian::le32_to_cpu(saveData->agentData.selectedMission);
+		saveData->agentData.nextMission = TFE_Endian::le32_to_cpu(saveData->agentData.nextMission);
+		for (s32 i = 0; i < 140; i++)
+		{
+			saveData->ammo[i] = TFE_Endian::le32_to_cpu(saveData->ammo[i]);
+		}
```
mlauss2, line 459: Since it's written to file, shouldn't these all be "cpu_to_le32()"  ?

mlauss2:
I compared the disassembly in a few places and at least on x86 all your changes are simply optimized away by the compiler, so good job :)
What system did you test this on? Are the saves of BE and LE compatible?

TheForceEngine/TFE_DarkForces/agent.cpp
BSzili: Sure thing, I'll add the reverse to the header for some syntactic sugar.

mlauss2: I just looked closer at this function: it's used for the darkpilo.cfg file which is created by TFE, therefore you don't need any endianness conversion.

BSzili: The reason I added byteswaps here so that people can continue where they left off in the dos version. It's also easy to copy this file along with the rest of the game data which can lead to confusion if it doesn't just work.

BSzili:
Just rebased this to the latest head. Let me know if there's anything else I should change/add before it can be merged.

PR #283, opened 2023-07-04
actor.cpp, flyers.cpp, sewer.cpp: replace `anim.state` magic numbers ...

kevinfoley:
…with explicit enum values


luciusDXL:
Looks good so far, I will merge once I spend more time making sure the numbers match. :)

TheForceEngine/TFE_DarkForces/Actor/actorModule.h
```diff
@@ -30,13 +30,20 @@ enum LogicAnimFlags
...
 	STATE_FIRE2,
 	STATE_ANIMATE2,
-	STATE_COUNT
+	STATE_COUNT,
+	//THINKER MODULE
+	// Move towards our destination
+	STATE_MOVE = 1u,
+	// Update our destination (generally moving towards the player).
+	STATE_TURN = 2u,
 };
```
jerethk, line 47: @kevinfoley My opinion is that it would be cleaner, and more readable code if we just make a separate enum

So change the existing `LogicAnimState` to `LogicAttackAnimState` (or something like that)

and add a `LogicThinkerAnimState` with MOVE and TURN

jerethk: as discussed in Discord - this won't work because both modules use the same struct

jerethk:
@kevinfoley looks like you also need to make a change in the `actor_thinkerModuleInit` method in `actor.cpp`
Currently it has `thinkerMod->anim.state = STATE_FIRE1;`
I guess that should be changed to "TURN"

luciusDXL (approved):

PR #285, opened 2023-07-05
Add setting to optionally fix DF `targetOffset` AI bug

kevinfoley:
There are two places in the Dark Forces AI code where the AI computes a `targetOffset` as part of its pathfinding. The original version of Dark Forces only considers the X axis when calculating the offset. As noted by lucius:

> This is obviously a typo and bug in the DOS code and should be `min(|dx|, |dz|)` but the original code is `min(|dx|, |dx|) => |dx|`

This pull request adds a new option to the TFE Game settings menu, "Fix target offset bug". When this option is enabled, the target offset is calculated from `min(|dx|, |dz|)` as was originally intended.

This has a subtle effect on AI movement. I tested with the 'trooper' units (stormtrooper, officer, commando) and found that the difference isn't noticeable without debugging aids, which is probably why the original team missed the bug. 

To help me understand what was going on under the hood and verify that the fix was working, I kludged together a debug visualization tool (not included in this PR) for comparing how the AI would move with or without the bug fix:

Target Offset

The white line shows the path that the AI would take without the bug fix, while the red line shows the path that the AI would take with the bug fix. This image shows a particularly extreme example; usually the difference between the two paths is less than 15 degrees. On the left side of the image, you can see past visualizations where the difference is much smaller.

Given that this is a subtle change, it may not be worth incorporating into TFE, but I figured I may as well make a PR.

### DIFF
--- a/TheForceEngine/TFE_DarkForces/Actor/actor.cpp
+++ b/TheForceEngine/TFE_DarkForces/Actor/actor.cpp
@@ -29,6 +29,7 @@
 #include <TFE_Jedi/Memory/list.h>
 #include <TFE_Jedi/Memory/allocator.h>
 #include <TFE_Jedi/Serialization/serialization.h>
+#include <TFE_Settings/settings.h>
 
 using namespace TFE_Jedi;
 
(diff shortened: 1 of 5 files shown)


luciusDXL:
I don't plan on taking this change, since it doesn't have a noticeable enough positive impact to warrant adding a new option. I might revisit this in the future once enemy modding is implemented.

PR #287, opened 2023-07-08
New debugging features

kevinfoley:
- New console command **`showMessage(message, [priority])`**: displays a message in the HUD, optionally with priority. 
- New CVAR **`d_showData`**: when enabled, shows an extended version of what you get from the LADATA cheat (adds pitch, yaw, sector, tick)
- New CVAR **`d_enableDebug`**: controls debug rendering
- `fixedPoint.h`: added `angleToFloatDegrees()`
- `console.cpp`: fixed messages overflowing the console if they contained newline characters


kevinfoley:
I used the same `d_` prefix for the CVARs as you've used for other CVARS. I'm guessing that the `d_` prefix stands for `display`, `draw`, or `debug`; in any case, it seems applicable. It might be helpful if you add a style guide for CVAR naming with other prefixes for future use (unless there's already one in the project that I've missed).

PR #289, opened 2023-07-17
VUE interpolation fixes

kevinfoley:
Also added two new distance functions to core_math.h


kevinfoley:
I found some additional issues with the Cargo VUE in Nar Shaddaa; I'll fix and then re-open the PR.

PR #299, opened 2023-08-01
Accessibility fixes

kevinfoley:
- Fix A11y system not distinguishing between captions and subtitles (by restoring code I had accidentally removed while refactoring)
- Caption/subtitle timing is now measured in microseconds instead of milliseconds to fix timing issues at very high framerates

### DIFF
--- a/TheForceEngine/TFE_A11y/accessibility.cpp
+++ b/TheForceEngine/TFE_A11y/accessibility.cpp
@@ -116,6 +116,7 @@ namespace TFE_A11Y  // a11y is industry slang for accessibility
 	void onSoundPlay(char* name, CaptionEnv env)
 	{
 		const TFE_Settings_A11y* settings = TFE_Settings::getA11ySettings();
+		// Don't add caption if captions are disabled for the current env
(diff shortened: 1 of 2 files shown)


TheForceEngine/TFE_A11y/accessibility.cpp
```diff
@@ -178,7 +198,7 @@ namespace TFE_A11Y  // a11y is industry slang for accessibility
 		Caption caption;
 		caption.text = args[1];
 		caption.env = CC_Cutscene;
-		caption.msRemaining = caption.text.length() * 90 + 750;
+		caption.microsecondsRemaining = caption.text.length() * 90 + 750;
```
luciusDXL, line 201: Does this calculation need to be updated since you are changing from Milliseconds to Microseconds?

kevinfoley: @luciusDXL D'oh! Yes, I missed that function; it's only used when displaying a test caption from the console. I can fix it later today.

luciusDXL (approved):

PR #308, opened 2023-08-08
Accessibility: added Unicode support, language and font dropdown menus, additional captions, logging CVAR, fixes

kevinfoley:
- Added a new "Subtitle/caption file" dropdown to the Accessibility menu. It looks for files in the `Captions/` directory (both in the TFE program directory and the TFE user documents directory) with file names in format `subtitles-XX.txt`, where XX is intended to be an ISO 639-1 language code (but currently can be anything). The dropdown just displays the file names for now.
_(Note: this screenshot shows an "es" file for the sake of example, but we do not actually have a Spanish subtitles file.)_
Subtitles file dropdown
- Added a new "Font" dropdown to the Accessibility menu. It looks for files in the `Fonts/` directory (also in both the program directory and user docs directory).
- The caption system now supports Unicode characters with UTF-8 (when used with a compatible font)
tfe_gif_Mon-Sep-18-21_43_40-2023_6
- Added missing subtitles for Kyle's radio VO on M11 Imperial City
- Added CVAR `d_logSFXNames` (bool). When enabled, the name of each sound effect that plays is logged to the console. This is very helpful when you want to caption a sound effect but don't know the name of the effect.
- Added captions for death SFX for most enemies


RC2225:
You probably should use TFE_PATHS::getPath() for populating the list with the files. At least on Linux it would not work without even when the file is in your working directory. 

Something like this. (Copied from the ModLoader) 
```
char captionsDir[TFE_MAX_PATH];
snprintf(captionsDir, TFE_MAX_PATH, "%sCaptions/", TFE_Paths::getPath(PATH_PROGRAM));
FileList dirList;
FileUtil::readDirectory(captionsDir, "txt", dirList);
		
```

kevinfoley:
@RC2225 I've updated it to search for subtitle files in the way that you suggested. I'm not sure exactly how #309 relates to this or how to fix it. Do the subtitles work on Linux in the current release (i.e. before I added the new dropdown?)

luciusDXL:
Thanks for the PR, I will go through it later today.

RC2225:
> @RC2225 I've updated it to search for subtitle files in the way that you suggested. I'm not sure exactly how #309 relates to this or how to fix it. Do the subtitles work on Linux in the current release (i.e. before I added the new dropdown?)

Sorry for the confusion, was more a remark what might not work with what I copied / don't fully understand. Furthermore, yes, before the dropdown was added it worked, which I completely forgot to mention. I also tested it now with the old version and when installed in the corresponding folders, which I before never did. With the old version it works too, even when the binary and the "support data" is not in the same folder. I see that I  still have to learn much about how native (linux) apps works. :)

TheForceEngine/TFE_A11y/accessibility.cpp
```diff
@@ -65,11 +75,98 @@ namespace TFE_A11Y  // a11y is industry slang for accessibility
...
+		assert(s_status == CC_NOT_LOADED);
 		CCMD("showCaption", addCaption, 1, "Display a test caption. Example: showCaption \"Hello, world\"");
+		CVAR_BOOL(s_logSFXNames, "d_logSFXNames", CVFLAG_DO_NOT_SERIALIZE, "If enabled, log the name of each sound effect that plays.");
+
+		// Get all caption file names from the Captions directories; we will use this to populate the
+		// dropdown in the Accessibility settings menu.
+		// First we check the User Documents directory for custom subtitle files added by the user.
+		char docsCaptionsDir[TFE_MAX_PATH];
+		const char* docsDir = TFE_Paths::getPath(PATH_USER_DOCUMENTS);
+		sprintf(docsCaptionsDir, "%sCaptions/", docsDir);
```
mlauss2, line 90: For better linux compatibility, please add
TFE_Paths::mapSystemPath(docsCaptionsDir);
in here (it's a noop on windows).

TheForceEngine/TFE_A11y/accessibility.cpp
```diff
@@ -65,11 +75,98 @@ namespace TFE_A11Y  // a11y is industry slang for accessibility
...
+		findCaptionFiles(docsCaptionsDir);
+		// Then we check the Program directory for subtitle files that shipped with TFE
+		char programCaptionsDir[TFE_MAX_PATH];
+		const char* programDir = TFE_Paths::getPath(PATH_PROGRAM);
+		sprintf(programCaptionsDir, "%sCaptions/", programDir);
```
mlauss2, line 95: same here

kevinfoley:
@mlauss2 I received several emails about comments you'd left on the code, but the links in the emails don't work and I don't have any notifications on GitHub; did you retract your comments?

mlauss2:
> @mlauss2 I received several emails about comments you'd left on the code, but the links in the emails don't work and I don't have any notifications on GitHub; did you retract your comments?

Yes, I commented on the subtitle paths wrt. Linux; but after looking closer again, the comments were wrong/too simplistic.

luciusDXL (approved):

luciusDXL (changes requested):
It looks like filePathList.* is in the root directory along with main. It probably makes sense to either put it in A11Y since that is the namespace you use.

luciusDXL (approved):

Issue #319, opened 2023-09-08
error compiling in file TheForceEngine/TheForceEngine/TFE_Jedi/Level/levelBin.cpp (Linux)


orderorder:
I had to add  #include <string.h> to solve the error.

luciusDXL:
Fixed in the 1.09.5 Release.

PR #323, opened 2023-09-13
Fix the build process on GNU/Linux systems.

maxz:
The cstring header was missing, therefore memcpy, memset, strcasecmp, strcat and strcpy could not be found during compilation.


maxz:
Resolves #319

luciusDXL (approved):

Issue #327, opened 2023-09-22
Stack Smashing in Linux Endeavour OS


orderorder:
[Startup] TFE_System::init
[Display] Fullscreen enabled.
[Display] Vertical Sync enabled.
[RenderBackend] OpenGL Device Tier: 3
[Startup] TFE_AudioSystem::init
[Audio] Attempting to initialize Audio System using API 'Linux ALSA'.
[Audio] Default Audio Output ID: 0.
[Audio] Device Count: 10.
[Audio] Device: 'default', ID: 0, Output Channel Count: 64.
[Audio] Device: 'hw:HDA ATI HDMI,3', ID: 1, Output Channel Count: 8.
[Audio] Device: 'hw:HDA ATI HDMI,7', ID: 2, Output Channel Count: 8.
[Audio] Device: 'hw:HDA ATI HDMI,8', ID: 3, Output Channel Count: 8.
[Audio] Device: 'hw:HDA ATI HDMI,9', ID: 4, Output Channel Count: 8.
[Audio] Device: 'hw:HDA ATI HDMI,10', ID: 5, Output Channel Count: 8.
[Audio] Device: 'hw:HDA ATI HDMI,11', ID: 6, Output Channel Count: 2.
[Audio] Device: 'hw:HD-Audio Generic,0', ID: 7, Output Channel Count: 6.
[Audio] Device: 'hw:HD-Audio Generic,1', ID: 8, Output Channel Count: 2.
[Audio] Device: 'hw:HD-Audio Generic,2', ID: 9, Output Channel Count: 0.
[Audio] Audio initialized using API 'Linux ALSA'.
[Audio] Starting up audio stream for device 'default', id 0.
[Startup] TFE_MidiPlayer::init
[Startup] TFE_Polygon::init
[Startup] TFE_Image::init
[Startup] TFE_FrontEndUI::init
[MemoryRegion] Allocated new memory block in region 'game' - new size is 1 blocks, total size is '8388608'
[MemoryRegion] Allocated new memory block in region 'level' - new size is 1 blocks, total size is '8388608'
[Error : a11y] Couldn't find font file at ; falling back to default font
*** stack smashing detected ***: terminated
[Error : CrashHandler] Received Signal 6 errno 10149 code 0
[Error : CrashHandler] Backtrace 9:
[Error : CrashHandler] 000 ./theforceengine(+0x1ac419) [0x56484ebaa419]
[...]

Issue #326, opened 2023-09-22
#include <cstring> in screenDrawGPU.cpp


orderorder:
TheForceEngine/TFE_Jedi/Renderer/RClassic_GPU/screenDrawGPU.cpp

Need to add line :

#include <cstring>

maxz:
I just tried to confirm this issue but can't. There is no issue for me when compiling the current head `9859e62c35cec2d32da17da56487f7233fc575f5`.

orderorder:
Im compiling in Endeavour OS (Linux)

jrugia:
Attempting to compile this with GCC 11.3.0 on my Debian system produces same error.

Can confirm that adding '#include <cstring>' in 'TheForceEngine/TheForceEngine/TFE_Jedi/Renderer/RClassic_GPU/screenDrawGPU.cpp' fixes it and compiles fine.

maxz:
I could now also confirm it with GCC 12.3.1 on Gentoo.

classilla:
Also needed on Fedora.

luciusDXL:
Thanks for the report, I wonder why this isn't always consistent. Anyway, easy fix. :)

luciusDXL:
Fixed in master.

PR #331, opened 2023-09-24
Fix linux build

maxz:
The cstring header was missing in `screenDrawGPU.cpp` so the compiler could not find the `memcpy` function and there was a possible format string bug waiting in `accessibility.cpp`.

### DIFF
--- a/TheForceEngine/TFE_A11y/accessibility.cpp
+++ b/TheForceEngine/TFE_A11y/accessibility.cpp
@@ -158,7 +158,7 @@ namespace TFE_A11Y  // a11y is industry slang for accessibility
 	void loadDefaultFont(bool clearAtlas)
 	{
 		char fontpath[TFE_MAX_PATH];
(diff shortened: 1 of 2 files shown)


maxz:
Possibly also fixes #327, although I'm not sure about that yet.

devusb:
This fixed #326 for me on NixOS, but I still get #327 when trying to run it.

TheForceEngine/TFE_A11y/accessibility.cpp
kevinfoley, line 1: Thanks for catching this; the same change will be necessary in `console.cpp` at line 67 (which is what I had used for reference).

kevinfoley: Disregard that comment - the string in `console.cpp` at line 67 is hardcoded, not a variable.

luciusDXL:
These changes were already made in different PRs/submits. So I'm closing the PR.

PR #335, opened 2023-09-26
Use SDL2 Mutexes and Threads

mlauss2:
Replace the TFE Threads and Mutexes with SDL2 equivalents.


luciusDXL (approved):
Thanks, this does remove a lot of Linux specific code, which is nice. :)

PR #336, opened 2023-09-27
A11y: Don't init captions if they are disabled; improved settings UI

kevinfoley:
- If none of the caption modes are enabled when TFE starts, we now skip the caption initialization process. Captions will be initialized if the user navigates to the Accessibility menu.
- Improved UI layout for Accessibility settings menu; fixed issue where the example caption was obscuring some settings on small screens.
- Restored functionality that adjusts the maximum number of caption lines during cutscenes based on the selected font size.

### DIFF
--- a/TheForceEngine/TFE_A11y/accessibility.cpp
+++ b/TheForceEngine/TFE_A11y/accessibility.cpp
@@ -354,7 +354,7 @@ namespace TFE_A11Y  // a11y is industry slang for accessibility
 		};
 
 		s_status = CC_LOADED;
-		delete s_captionsBuffer;
+		free(s_captionsBuffer);
 	}
 
 	void onFileError(const string path)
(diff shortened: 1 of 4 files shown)


kevinfoley:
@mlauss2 I can't believe I missed that typo! Thanks for pointing it out. I did add a divider and spacing as part of the UI improvements.

luciusDXL (approved):

TheForceEngine/TFE_A11y/accessibility.cpp
```diff
@@ -146,7 +154,9 @@ namespace TFE_A11Y  // a11y is industry slang for accessibility
 
 		// Try to load the previously selected font.
 		string lastFontPath = TFE_Settings::getA11ySettings()->lastFontPath;
-		tryLoadFont(lastFontPath, false);
+
+		if (lastFontPath, ImGui::GetIO().Fonts->Locked)	{ setPendingFont(lastFontPath);	}
```
mlauss2, line 158: that "lastFontPath," is that a copy-paste leftover?

kevinfoley: Looks like that was supposed to be a `&&` instead of a comma.

mlauss2: like this?  (!lastFontPath.empty() && ImGui::GetIO.... ) ?

PR #339, opened 2023-09-27
Fix tiny mistakes in german subs

mlauss2:
"MIENE" (countenance) to "MINE" (mine/pit), this irritated me a lot in Dark Tide 2.

I also changed "Unterbruch" since this is (I think) a Swiss German expression, not commonly heard in Germany or Austria (to "Unterbrechung")

### DIFF
--- a/TheForceEngine/Captions/subtitles-de.txt
+++ b/TheForceEngine/Captions/subtitles-de.txt
@@ -20,9 +20,9 @@ m16moc01.voc "Offizier Mohc: Es ist lange her, seit ich einen Mann zum Kampf her
 //SFX
 boba-1.voc	"[Fett lacht]"
 boba-3.voc	"[Fett grunzt]"
-boba-4.voc	"[Fett Todesschrei]"
+boba-4.voc	"[Fetts Todesschrei]"
 beep-01.voc "[Alarm]"
-beep-10.voc "[MIENE PIPST]"
+beep-10.voc "[MINE PIEPST]"
 fall.voc 	"[Kyle schreit]"
 door.voc 	"[Türe zischt]"
 door-04.voc "[Türe bewegt sich]"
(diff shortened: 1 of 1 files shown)


RC2225:
Thanks. Ich ziehe eine Miene, dass ich Mine mit IE geschrieben habe. In the line with Unterbrechung, maybe there is even a better word the English subs dont even use something like this in this line. IIRC the german line is more understandable than the english line.

RC2225:
Störung sounds good. If you are used to MW Radio I would even say this message has a very clean signal. 

M07JAN02: its Wo instead of wohin. 
M16KYL04: its Jan würde stolz auf mich sein.
M06JAN01: Its beeil and dass
m01mma04: Sehr richtig Commander. Ich hoffe Sie werden verstehen, dass alles was wir hier besprechen vertraulich ist
m01mma05: Einem neuen Typ

maxz:
Looks fine.

I'm not a huge fan of "Fetts Todesschrei", but can't come up with anything better that is as succinct either.

RC2225:
Probably just schreit is sufficient. I used Todesschrei that its clear and a bit because I thought its "funny". But I'm not really happy either. For me it sounds like between the crying in a coming like: Buuhäää

maxz:
I would celebrate something like "Fett haucht sein Leben aus", although that would probably not be dramatic enough.

RC2225:
I mean canonically he cannot die. The English subs use screams. 

m01mma04 there is a missing ,  before dass.  I would probably add another after alles. 
fun fact M16KYL04 its in the sound files its really würde, but wäre sounds better. IIRC it was once mentioned in the discord.

maxz:
I did not even remember that DF had German audio. My recent playthrough was in English. Then it obviously makes sense to stick to the audio.

I guess I must have played the German version back in the 90s then, because if there was one I surely did not play the English one.

RC2225:
On Steam you have the English, German, French and Spanish version. Just change it in the steam game settings. For the German version is patch since two voice lines are not translated in the release version. Furthermore, there was a Korean release.

luciusDXL (approved):

PR #340, opened 2023-09-27
SDL_Audio support.

mlauss2:
Replace RtAudio with SDL2-Audio system.  Requires at least SDL 2.24 (which is the minimum on Linux).
EDIT: with patch 2 minimum SDL2 is now 2.0.6.


luciusDXL:
This actually looks pretty good, but can you fix the merge conflict?

luciusDXL:
I will need to test locally later, but assuming that succeeds I will merge the changes. Thanks.

luciusDXL (approved):

PR #341, opened 2023-09-27
Linux/CMake: make RtMidi optional

mlauss2:
If RtMidi library is not found on build system, the System Midi music device will not be build (SF2 and OPL3 will still be available) and the final binary will not require RtMidi at all.

This could be useful for flatpatk/snap and SteamDeck builds since it gets rid of an external dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,9 @@ else()
 endif()
 include(GNUInstallDirs)
 
+## Options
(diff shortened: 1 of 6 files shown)


luciusDXL:
Please resolve the conflicts and then I can accept it. Thanks.

luciusDXL (approved):

luciusDXL:
Ok, looks good. Thanks.

PR #345, opened 2023-09-29
UI_Images: fix up png formats

mlauss2:
Found while playing with SDL-image. These 2 images are not like the others, i.e. not 32bit.  If applied, a conversion step to the TFE-expected 32bit pixel format can be avoided.

TFE_LoadSelected.png: convert from 8bit to 32bit
TFE_TitleLogo.png: increase width from 321 to 322 pixels, convert to 32bit


luciusDXL (approved):

luciusDXL:
Strange, thanks for the fix.

PR #346, opened 2023-09-29
MidiPlayer: select OPL3 as default midi device

mlauss2:
if an invalid midi-system ID is passed in (e.g. via config file), TFE only complains about it, but this will lead to a crash as soon as the midi device is queried (either Sound settings page or game).

Install the OPL3 device as fallback.


luciusDXL (approved):

luciusDXL:
Fair enough.

PR #349, opened 2023-09-30
Editor: Fix up on Linux

mlauss2:
Enable the Editor on Linux. Adds CMakeFiles and compiler fixes (missing headers mostly).
It runs, doesn't crash, but I don't see a lot in the AssetBrowser, and the LevelEditor also shows no level geometry.


luciusDXL:
I had originally just disabled the code on Linux so I could go back and figure out paths etc. later. You should see textures, sprites, frames, and other assets in the Asset Browser. As for the level editor, its expected that you don't see anything yet - its basically just stubbed out at the moment (not yet restored).

luciusDXL (approved):

luciusDXL:
I went ahead and took the changes - they were going to need to happen eventually anyway. Thanks. :)  At some point I will have to figure out why you are not seeing the textures and other assets though.

PR #354, opened 2023-10-01
MIDI fixes

mlauss2:
- Bring back the midiPlayer::suspend/resumeThread() functions.  While POSIX cannot really suspend a thread like windows can, we can effectively block the midi thread by acquiring the lock: this way the thread can complete it's work and won't run until the lock is released.  Should fix lucius' windows crash when switching the midi device.

- introduce a "(Disabled)" device at index 0: RtMidi tends to spam the log with error messages about writing to non-existing ports. This device is always present and moves all bits to /dev/null.
One can then add and remove midi synthesizers on the fly and they'll show up in the settings screen.


luciusDXL (approved):

luciusDXL:
Alright, we'll try this and see if the crash returns. Thanks for looking at it again.

luciusDXL:
Ok, that appears to fix the crash. Thanks.

Issue #310, opened 2023-08-08
[Bug/Linux] SoundFonts directory handling, log spam when not detected


SeongGino:
A two-piece issue, both stemming from the same problem.
TFE loads SoundFonts at a hardcoded directory, `$HOME/SoundFonts`, which might produce home folder clutter for users. Perhaps we should be using `$XDG_DATA_HOME/TheForceEngine/SoundFonts` (i.e. `~/.local/share`) as we're already using for the configuration files and other user data?

This folder is not made by TFE, but creating one does allow SF2 soundfonts to work as expected - but again, not ideal to dump unneeded folders in the home dir.
___

On a related note, if TFE doesn't detect this folder, as soon as the audio engine is booted (and *"SF2 Synthesized MIDI"* is selected as the currently active Midi device), the log is spammed with:

```
...
[Error : readDirectory] opendir(/home/user/SoundFonts/) failed with 2

[Error : readDirectory] opendir(/home/user/SoundFonts/) failed with 2

[Error : readDirectory] opendir(/home/user/SoundFonts/) failed with 2

[Error : readDirectory] opendir(/home/user/SoundFonts/) failed with 2
...
```
(Posted a log file for demonstration: [1.6mb large for sitting at the title screen for maybe all of six seconds. Ouch.](https://github.com/luciusDXL/TheForceEngine/files/12296575/the_force_engine_log.txt))

mlauss2:
Yes, the same issue as #309

SeongGino:
The log spam seems to have been resolved and TFE is reading most of the data files from `/usr/share/TheForceEngine`. I also put a mod zip into `~/.local/share/TheForceEngine/Mods` (had to make the mods folder) and it reads that fine.

I noticed though that the soundfonts only seem to be read from the /usr/share directory now, which is read only on virtually all Linux distros. Is this intentional? For normal use, this is fine, but that does mean using custom SF2s would be more difficult (or at least require root access, which may be seen as needless if they just want a different soundfont and rather not deal with system MIDI).

PR #355, opened 2023-10-03
[RFC] new path handling prototype

mlauss2:
Based on comment in #310, I implemented (on linux for now) a new scheme for finding TFE Support files
like SoundFonts, Captions, Shaders, ...

You request the function to return all files from subfolder e.g. "SoundFonts", with extension ".sf2", and
you get back a vector of tuples which contain, in order of path preference, a list of  paths, filenames  and extension.  The function makes sure to read all known paths (PATH_PROGRAM with highest priority, PATH_PROGRAM_DATA with 2nd prio, PATH_USER_DOCUMENTS) with least priority.  The priority determines which file to use in case the name exists in more than one directory.  No file is returned more than once (based on the name of the file).

I've implemented this in the soundfont2 device; and symlinked the user_data soundfont folder to the system soundfont cache:
image

I think this could also be used for Captions+Fonts (which uses a similar system filePathList) and Mods.
I've opened this PR just for discussing the idea, so please ignore the fileutil-posix.c changes for now.

### DIFF
--- a/TheForceEngine/TFE_FileSystem/fileutil-posix.cpp
+++ b/TheForceEngine/TFE_FileSystem/fileutil-posix.cpp
@@ -1,3 +1,7 @@
+#include <algorithm>
+#include <tuple>
+#include <utility>
+#include <map>
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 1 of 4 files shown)


TheForceEngine/TFE_FileSystem/fileutil-posix.cpp
```diff
@@ -19,6 +23,99 @@ namespace FileUtil
 	bool existsNoCase(const char *filename);
 	static char *findFileObjectNoCase(const char *filename, bool objisdir);
 
+	// paths to gather files from, in descending order of priority.
+	static const int path_prio[] = { PATH_USER_DOCUMENTS,
```
luciusDXL, line 27: It looks like PATH_USER_DOCUMENTS is repeated twice.

luciusDXL:
I think the names need some work (FilePath can refer to a proper path or as part of an archive). I think overall its fine if we want users to be able to put files in all of these locations.

Though I'm still a bit confused on how installing TFE on Linux results in all of these different paths, instead of always putting the data in the same (relative) place (since it is data that ships with the releases) - I wonder if that can be fixed somehow instead.

kevinfoley:
@luciusDXL For captions and fonts, I support user-added captions and fonts, which can be installed in either the program directory or the documents directory. Other than that, I am somewhat unclear about the directory issues on Linux because I am primarily a Windows user.

I did set up a Fedora 38/KDE dev environment in the hopes that I can replicate and better understand the path issues on my end, but after building TFE I can't get it to run (it seems like it can't find my GPU; I need to ask about that on the Discord).

TheForceEngine/TFE_FileSystem/fileutil.h
kevinfoley, line 1: Perhaps it would be better to use a struct with named fields, rather than a tuple with three helper functions?

PR #359, opened 2023-10-11
Added A11y setting to disable fullscreen flashes

kevinfoley:
Adds a setting to disable fullscreen flashes, e.g. when the player takes damage or picks up an item. This may be helpful for players with photosensitive epilepsy.


TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
```diff
@@ -2799,6 +2799,14 @@ namespace TFE_FrontEndUI
 		ImGui::LabelText("##ConfigLabel4", "Motion Sickness");
 		ImGui::PopFont();
 		ImGui::Checkbox("Enable headwave", &a11ySettings->enableHeadwave);
+
+		// OTHER -----------------------------------------------
+		ImGui::Dummy(ImVec2(0.0f, 10.0f));
```
luciusDXL, line 2804: What does this do?

luciusDXL (approved):

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
kevinfoley: @luciusDXL That is apparently how you add empty space of a specific size to an ImGui layout. I had to look it up, no way I ever would have guessed that on my own. So there I'm adding 10 units of vertical space.

luciusDXL: Ok, thanks.

PR #361, opened 2023-10-13
Use mac data

### LINKED ISSUE #77: Support Mac Data
Add support for higher resolution Mac data.

stephengeorgewest:
fixes #77 
I think this covers all the asset differences:

- [x] 640x480 loading screen
- [x] double resolution hud textures
- [x] Cutscenes" folder instead of "LFD" folder
   - [x] agent menu button says "Quit" instead of "Dos", just works no
   - [x] agent menu hotkeys D shortcut should changes to Q
   - [ ] confirm quit menu still says "Dos", but there are no asset changes for that. Mac didn't use that ui, it used native dialogs. Maybe we will need a mod for that.
   - [ ] agent menu/mission briefing on mac is rendered at 640x480, and seems to be upscaled somehow, since the brf-(mon/jan/nil) animations are identical, and the font is nearly identical. I think the differences are in the mac binary doing the upscaling.
   - [ ] ? FONT8 was modified slightly, I didn't have a good way to view the changes, but I didn't notice any problems, so I assume it was used without issue?
 - [x] Extra gob `WEAPONS.GOB` which holds weapons images, which are identical to the weapons images in dos `TEXTURES.GOB`
 - [x] TEXT.MSG changes work fine. but "command-n" isn't really a viable shortcut? ESC still brings up the menu with next mission
 - [x] changes to levels seem to work just fine
   - [x] `SECBASE.O`
     - [x] animation on death star plans
   - [ ] `NARSHADA.O'
     - [x] animation on nava card
     - [ ] ? missing item, remote, I assume is missing, I didn't check which one it was. it is at `X: -23.08 Y: -10.00 Z:-168.08` if anybody has a way to check where that is
   - [ ] ? fuelstat, I wasn't sure how to check these, but a quick run-through of the level seemed to work ok.
     - [ ] sequences for sector `hockabout2` changed a bit, messages (`arm1shell` `arm2` `arm2shell` `farkearm2`) `next_stop` were added
     - [ ] changes for re-ordered arm1 <-> arm2 messages changed, but it looks the same content

Questions/out of scope
 - when loading `WEAPONS.GOB` I didn't see a way to error if it is missing, but not error if we don't need it? I just changed it to a warning and updated the message so dos users can know it isn't really required for them.
 - mac rendered at 640x400, or similar wide-ish screen formats, but placed the hud in a black section underneath. I don't know if anybody wants that.
 - mac also offered 320x240 doubled pixels, 320x240 interlaced and some other sizes, but they were all in a black box, or doubled no other streaching. Rendere would need some other changes for the true "mac experience".
 - gpu render scales the hud down to match the size of the smaller assets, the extra bits on the side for 4:3 hud on widescreen look odd being twice as pixely.
 - software renderer doubles the pixels, so at 320x200 it looks big, but at 640x480 it looks normal. but continuing to scale up the hud gets smaller, which seems to just match what happened before. side bits don't match as they are half sized now.

### DIFF
--- a/TheForceEngine/TFE_DarkForces/GameUI/agentMenu.cpp
+++ b/TheForceEngine/TFE_DarkForces/GameUI/agentMenu.cpp
@@ -245,7 +245,7 @@ namespace TFE_DarkForces
 				s_buttonPressed = AGENT_REMOVE;
 				s_buttonHover = JTRUE;
 			}
-			else if (TFE_Input::keyPressed(KEY_D))	// DOS
+			else if (TFE_Input::keyPressed(s_langKeys->k_quit))	// DOS
 			{
 				s_buttonPressed = AGENT_EXIT;
 				s_buttonHover = JTRUE;


classilla:
Excellent!

PR #364, opened 2023-10-19
Added a Portable mode

kevinfoley:
Added support for running TFE in "Portable Mode", where temporary files and user documents are stored in subfolders inside the executable directory. Portable mode can be activated in either of two ways:

- By launching the executable with the new command line argument `-portable`
- By adding a file named `PORTABLE` to the executable directory

User documents are saved within a new directory named `UserDocs` inside of the executable directory.

Tested on Windows 10 and Fedora 38/KDE Plasma


luciusDXL:
I have to go through this one again, but I will do so in the next few days.

luciusDXL:
> just bikeshedding here:
> 
> * Maybe add an envvar too?  i.e. "TFE_PORTABLE=1 ./theforceengine",
> * Maybe make it configurable as default behaviour (thinking about a flatpak here again where everyhing the needs is where the exe is, like windows).
> 
> Hm?

I'm not a fan of using an environment variable on Windows. That said, making it configurable via CMake sounds like a great idea for Linux.

Tartifless:
Avoid environment variable, argument or presence of a portable.txt file is the way to go

Tartifless:
Hello, any news regarding this feature, are there any plans to integrate it in the next release ?

kevinfoley:
> Hello, any news regarding this feature, are there any plans to integrate it in the next release ?

@luciusDXL ?

luciusDXL:
I need to go through this again and either merge it or implement my own version. It won't be in the next release, but should follow soon after.

luciusDXL:
I was investigating problems with Documents/ on Windows - mainly due to OneDrive - and realized that on failure the system was already supposed to use the local directory. That wasn't quite happening (not checking in all of the correct places), but with that in place it made sense to directly support portable installs.

So I basically implemented it, then remembered this PR. My version is similar but with some notable differences:
* There is no commandline option or env variable, since most Windows users wouldn't use it.
* Instead of creating a new file, you get the "portable" behavior by placing a settings.ini file in your TFE directory (if it is empty, it treats it like a first run, but you can safely copy the existing data as well) - this is similar to the way GZDoom handles it.
* It doesn't create another sub-directory, most people wouldn't link "MyDocs" to settings, saves, logging, etc.

Thanks for taking the time to implement this @kevinfoley. I'm going to close it now.

Tartifless:
Hello Lucius, will this be implemented in next version ?

luciusDXL:
I have not released a new version with this code yet, it is work that will be included in the next release.

Tartifless:
Hello, thanks, indeed I figured that out by looking at the commit history.
Anyway, very much appreciated

PR #365, opened 2023-10-19
Added Dark Forces cheat menu

kevinfoley:
Also added new cheat LACAT - gives you 9 lives.

Cheat Menu 2


luciusDXL (approved):
Looks good, thanks for the contribution.

PR #367, opened 2023-10-23
add a Ubuntu build workflow for github

mlauss2:
Adds a github workflow file to build the master branch on every push.

### DIFF
--- a/.github/workflows/ubuntu-gcc.yml
+++ b/.github/workflows/ubuntu-gcc.yml
@@ -1,0 +1,36 @@
+# This starter workflow is for a CMake project running on a single platform. There is a different starter workflow if you need cross-platform coverage.
+# See: https://github.com/actions/starter-workflows/blob/main/ci/cmake-multi-platform.yml
+name: Linux/x64 Test-Build master branch
+
+on:
+  push:
+    branches: [ "master" ]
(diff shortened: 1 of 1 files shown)


luciusDXL (approved):

luciusDXL:
The workflow fails due to SDL - but I will just leave it for now due to lack of time.

luciusDXL:
@mlauss2 Alright, that seems to work. Thanks for setting up the workflow, that will help avoid accidentally breaking Linux compilation on my end.

luciusDXL:
So the plan is to enable building the editor since that is the current focus (I see the scripting stuff is already enabled, which is great). And in general, thanks for all of the help - especially with Linux. Especially putting up with my Windows-centric development. :)-

luciusDXL:
The editor was enabled and the build still passes. Thanks again.

luciusDXL:
> at some point in the future, once I have figured out how to actually do that, I'd like to enable automatic building of flatpaks when you cut a release.

That would be really great, honestly.

> This workflow won't catch everything though since the default compiler on ubuntu is fairly old. Most of the "bugs" I encounter are due to gcc-14's stricter implementation of c++.

Fair enough, but just catching things like missing headers will be nice.

PR #381, opened 2024-02-28
Update AppStream meta data

JakeSmarter:
* Rename file to match the app’s component id (sync to Flathub release) that conforms to AppStream spec
* Drop deprecated `application` element in favor of the `component` element with an appropriate `type` attribute
* Drop deprecated `developer_name` element in favor of the `developer` element
* Add more `url` types
* Add `suggests`, `recommends`, `provides`, and `control` elements
* Sync AppStream’s `categories` with Desktop Entry’s `Categories`
* Add de, en-US, and pl translations

See also flathub/io.github.theforceengine.tfe#3 and flathub/io.github.theforceengine.tfe#8.

FYI: Personally, I would have gone with `io.github.TheForceEngine` for the component id because it would have resembled the spec’s intention much closer. But, with Flathub’s already existing install base we would not want to “open a new can of worms”.


fpiesche:
> FYI: Personally, I would have gone with `io.github.TheForceEngine` for the component id because it would have resembled the spec’s intention much closer. But, with Flathub’s already existing install base we would not want to “open a new can of worms”.

I actually initially submitted my PR for the Flathub build with that id, but as `github.io` is not owned by the project, Flathub's reviewers requested I add a fourth component to the id to distinguish it from any other projects that may be under the `TheForceEngine` org.

luciusDXL (approved):

PR #386, opened 2024-03-01
Added workaround for GLEW on wayland

Rosalie241:
This patch adds a workaround to make GLEW initialize successfully on wayland, allowing the force engine to work properly when using wayland on linux, more information and existing workarounds can be found on the following issue:

https://github.com/nigels-com/glew/issues/172

I'm not sure if this kind of workarounds are desired, so feel free to close this PR if you have no desire for such a thing!


luciusDXL (approved):
Seems ok, but be sure to remove when/if this workaround becomes unnecessary.

fpiesche:
> @fpiesche This may enable or require adding `--socket=wayland` and `--socket=fallback-x11` in the flatpak build manifest.

As I understand from review feedback on projects, these are only required for applications that are actually built for Wayland so we probably shouldn't need either, unless this comes with TFE being built against Wayland rather than X11. I'll give it a try locally though!

PR #396, opened 2024-03-22
Change function name

jerethk:
Change function name from sector_canWallMove to sector_isWallBlockedByPlayer

I assume the name is one that you chose during RE and didn't come from the actual LEC code
I was just finding it a bit difficult to reason about what the function does with its existing name :-) 

An even more accurate name would probably be isMovingWallBlockedByPlayer (or isWallMovementBlockedByPlayer) but that's getting a bit long, happy for you to decide!


jerethk:
or maybe even `doesMovingWallCollideWithPlayer` ?

luciusDXL (approved):

PR #398, opened 2024-03-29
New optional setting: observe "solid wall" flag when testing if object overlaps a wall

jerethk:
This will cause the player to collide with a "solid" adjoined wall that is moving or rotating
Results in a fix for maps like Harkov and Dark Tide 3 which feature a morph_move or morph_spin elevator surrounded by "solid" adjoined walls to simulate riding in a vehicle

### DIFF
--- a/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
+++ b/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
@@ -3129,6 +3129,7 @@ namespace TFE_FrontEndUI
 				gameSettings->df_bobaFettFacePlayer = true;
 				gameSettings->df_smoothVUEs = true;
 				gameSettings->df_pitchLimit = (temp == TEMPLATE_MODERN) ? PITCH_MAXIMUM : PITCH_VANILLA_PLUS;
+				gameSettings->df_solidWallFlagFix = true;
 				// Graphics
 				graphicsSettings->rendererIndex = RENDERER_HARDWARE;
 				graphicsSettings->skyMode = SKYMODE_CYLINDER;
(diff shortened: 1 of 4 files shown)


TheForceEngine/TFE_Settings/settings.h
```diff
@@ -205,6 +205,7 @@ struct TFE_Settings_Game
 	bool df_autorun = false;			// Run by default instead of walk.
 	bool df_crouchToggle = false;		// Use toggle instead of hold for crouch.
 	bool df_ignoreInfLimit = true;		// Ignore the vanilla INF limit.
+	bool df_solidWallFlagFix = false;	// Solid wall flag is enforced for collision with moving walls.
```
luciusDXL, line 208: If this doesn't break any mods, I would suggest defaulting it to true so people get the best experience.

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
```diff
@@ -3171,6 +3177,7 @@ namespace TFE_FrontEndUI
 				gameSettings->df_showSecretFoundMsg = false;
 				gameSettings->df_bobaFettFacePlayer = false;
 				gameSettings->df_smoothVUEs = false;
+				gameSettings->df_solidWallFlagFix = false;
```
luciusDXL, line 3180: Same as the other comment, default to true if it doesn't break mods so people get the best experience. Most people won't ever change these settings.

TheForceEngine/TFE_Settings/settings.h
jerethk: Done. I cannot think of how it could break any mods...

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
jerethk: What I've done here is leave it as `false` for vanilla mode, and set it to `true` in modern mode

luciusDXL: Sounds good, thanks.

luciusDXL (approved):

PR #399, opened 2024-03-30
New Dev panel with lighting controls

kevinfoley:
Added a new Developer panel to the settings menu, which optionally exposes new controls to dynamically adjust lighting for 3DOs.

The Jedi engine uses 3 directional lights to illuminate 3DO models; however, the position and brightness of each light was hardcoded in the original engine. The lighting controls in the Developer panel allow you to adjust each of the 3 directional lights individually, exposing a capability that was built into the original software engine even back in 1995 but never properly utilized until now.

The Developer panel also includes an optional setting to import all 3DOs with gouraud shading, overriding the shading flags defined in the 3DO. This setting takes effect when models are loaded and therefore requires restarting the engine.
tfe_gif_Sat-Mar-30-00_02_53-2024_1
tfe_gif_Sat-Mar-30-00_22_07-2024_0

Currently the lighting settings in the developer panel only support software rendering.


luciusDXL:
I plan on integrating this PR after the next release (any day now).

luciusDXL (approved):

Issue #402, opened 2024-04-16
More CPU activity (even with OpenGL renderer) than dosbox


MIvanchev:
Hi there, I wanted to replay the DF campaign and figured out the best way to go ahead would be TheForceEngine (great project btw). However on my Thinkpad with iGPU it runs (in every configuration) slower than DF in DosBox. Even with the OpenGL renderer I experience 50% CPU activity and the fan doesn't stop blowing where as dosbox never turns the fan on. Any advice on this? I'm playing on Linux.

MIvanchev:
Did some research, the CPU usage spikes immediately after the program is launched, on the 2D menu. `top` shows 3 processes, 1 with very high CPU activity and 2 with ~5%.

Gerwin2k:
It was told somewhere in this repo that VSync enable would also schedule CPU-idling. But that did not work for my WIN32 environment. So a few months ago I hacked up this bit below , that works OK for the time being. Should be cross-platform.
Result is a total of 30% use of 4-core CPU, spread evenly across the four cores.

file: TFE_System/system.cpp

```
	void sleep(u32 sleepDeltaMS)
	{
		if (sleepDeltaMS==0) SDL_Delay(1);
		else SDL_Delay(sleepDeltaMS); 
	}
```

MIvanchev:
Yes, you appear to be correct, the main loop of TFE seems to do busy waits so the CPU is maxed out all the time... This seems to be a common SDL issue. Maybe a long term solution would involve SDL_WaitEventTimeout.

luciusDXL:
So you mean that it busy waits on your system even with the frame limiter enabled. The idea is to avoid wasting too much time but to give other threads a chance to run. But Windows has always been a bit iffy here - so I will try out your idea and see how well it works.

luciusDXL:
So it turns out using the 1ms sleep reduced the accuracy of the frame limiter too much, but with some minor modifications I was able to restore it. The change has been submitted and will be in the next release.

MIvanchev:
@luciusDXL You also need to address the midi thread, it maxes out a thread as well and doesn't even do sleep(0) just loops continuously.

Gerwin2k:
Thanks luciusDXL. I will try it when I have some time to spare.

I was slightly involved with a similar adjustment in another game port, some months ago. 
There it was implemented like this:

https://github.com/LostArtefacts/TR1X/pull/1002/files

luciusDXL:
[...]

PR #405, opened 2024-04-20
Avoid busy waiting on the main and the midi threads.

MIvanchev:
Solves #402.

Hi, I would like to suggest the following changes that dramatically improve the CPU usage by avoiding busy waiting in the threads. For the main thread I suggest using SDL_WaitEventTimeout in the frame limiting which pauses the thread until events arrive because I believe it's important to process events immediately. So if your mouse is, say, 1200Hz you will not be limited to 60FPS.

For the MIDI thread I propose a command condition which also awakes the thread immediately. Otherwise the thread is paused until it's time to play another note.

For me this brings down CPU usage from 30% to 10% and the fan stays off the whole time.

### DIFF
--- a/TheForceEngine/TFE_Audio/midiPlayer.cpp
+++ b/TheForceEngine/TFE_Audio/midiPlayer.cpp
@@ -14,6 +14,7 @@
 #include <TFE_Audio/MidiSynth/fm4Opl3Device.h>
 #include <algorithm>
 #include <assert.h>
+#include <limits>
 
 #ifdef _WIN32
 #define WIN32_LEAN_AND_MEAN 1
@@ -39,6 +40,26 @@ namespace TFE_MidiPlayer
 	{
 		MidiPlayerCmd cmd;
 		f32 newVolume;
+
+		static inline MidiCmd pause()
+		{
+			return { MIDI_PAUSE };
+		}
+
+		static inline MidiCmd resume()
+		{
+			return { MIDI_RESUME };
+		}
+
+		static inline MidiCmd stop()
+		{
+			return { MIDI_STOP_NOTES };
+		}
+
+		static inline MidiCmd changeVol(f32 vol)
+		{
+			return { MIDI_CHANGE_VOL, vol };
+		}
 	};
 
 	enum { MAX_MIDI_CMD = 256 };
@@ -59,9 +80,10 @@ namespace TFE_MidiPlayer
 	static SDL_Thread* s_thread = nullptr;
 	static bool s_tPaused = false;
 
-	static atomic_bool s_runMusicThread;
+	static bool s_threadAlive = true;
 	static u8 s_channelSrcVolume[MIDI_CHANNEL_COUNT] = { 0 };
 	static SDL_mutex* s_mutex;
+	static SDL_cond* s_midiCmdCond;
 
 	static MidiDevice* s_midiDevice = nullptr;
 	static MidiCallback s_midiCallback = {};
@@ -107,6 +129,13 @@ namespace TFE_MidiPlayer
 			TFE_System::logWrite(LOG_ERROR, "Midi", "cannot initialize SDL mutex");
 			return false;
 		}
+		s_midiCmdCond = SDL_CreateCond();
+		if (!s_midiCmdCond)
+		{
+			SDL_DestroyMutex(s_mutex);
+			TFE_System::logWrite(LOG_ERROR, "Midi", "cannot initialize SDL condition");
+			return false;
+		}
 
 		allocateMidiDevice(type);
 
@@ -122,9 +151,6 @@ namespace TFE_MidiPlayer
 				}
 			}
 		}
-
-		s_runMusicThread.store(true);
-
 
 		s_thread = SDL_CreateThread(midiUpdateFunc, "TFE_MidiThread", nullptr);
 		if (!s_thread)
@@ -149,11 +175,15 @@ namespace TFE_MidiPlayer
 
 		TFE_System::logWrite(LOG_MSG, "MidiPlayer", "Shutdown");
 		// Destroy the thread before shutting down the Midi Device.
-		s_runMusicThread.store(false);
+		SDL_LockMutex(s_mutex);
+		s_threadAlive = false;
+		SDL_CondSignal(s_midiCmdCond);
+		SDL_UnlockMutex(s_mutex);
 		SDL_WaitThread(s_thread, &i);
 
 		delete s_midiDevice;
 
+		SDL_DestroyCond(s_midiCmdCond);
 		SDL_DestroyMutex(s_mutex);
 	}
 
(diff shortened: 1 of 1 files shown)


luciusDXL:
I have seen this and plan on going through it after the next release is out (any day now). Thanks for the suggestions and PR. :)

MIvanchev:
@mlauss2 Let me rebase, running your CPU at 110% long term is not a good idea!!!

MIvanchev:
@mlauss2 I've rebased.

TheForceEngine/TFE_Audio/midiPlayer.cpp
```diff
@@ -191,12 +221,17 @@ namespace TFE_MidiPlayer
 	//////////////////////////////////////////////////
 	// Command Buffer
 	//////////////////////////////////////////////////
-	MidiCmd* midiAllocCmd()
+	void tryQueueMidiCmd(MidiCmd cmd)
```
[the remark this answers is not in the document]
MIvanchev: ah true :+1:

TheForceEngine/TFE_Audio/midiPlayer.cpp
```diff
@@ -505,10 +521,27 @@ namespace TFE_MidiPlayer
...
 				// Check for hanging notes.
 				detectHangingNotes();
+
+				// Sleep until the next playback window to
+				// avoid a busy waiting scenario.
+				auto timeout = (s_midiCallback.timeStep - s_midiCallback.accumulator) * 1000;
+				if (!std::isfinite(timeout))
+				{
+					timeout = std::numeric_limits<decltype(timeout)>::max();
+				}
```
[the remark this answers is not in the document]
MIvanchev: The goal of this code is to avoid (the very unlikely) undefined behavior.

TheForceEngine/TFE_Audio/midiPlayer.cpp
```diff
@@ -505,10 +521,27 @@ namespace TFE_MidiPlayer
...
+				if (timeout > std::numeric_limits<Uint32>::max())
+				{
+					timeout = std::numeric_limits<Uint32>::max();;
+				}
```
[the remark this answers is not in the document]
MIvanchev: IMHO SDL_MUTEX_MAXWAIT is actually intended to be a flag, not a number, this is bad interface design on SDL's part. However you're on to something that the value is actually an int so let's change that :)

MIvanchev:
Thanks for all your testing!

>  When not moving mouse/using keyboard, the framerate is about 5% off, when moving the character it goes up to 120, and the whole thing is really choppy when target framerate is set to 30fps.

Yeah that limiter was not that good, let's kick it out.

mlauss2:
@MIvanchev please redo the patch with the frame limiter bits removed.  Thanks!

MIvanchev:
I haven't forgotten @mlauss2, just need some time to do it :) Let me get to it ASAP.

TheForceEngine/TFE_Audio/midiPlayer.cpp
```diff
@@ -505,10 +521,27 @@ namespace TFE_MidiPlayer
 
 				// Check for hanging notes.
 				detectHangingNotes();
+
+				// Sleep until the next playback window to
+				// avoid a busy waiting scenario.
+				auto timeout = (s_midiCallback.timeStep - s_midiCallback.accumulator) * 1000;
+				if (!std::isfinite(timeout))
```
mlauss2, line 528: This doesn't work with VisualStudio, std::isfinite() is unrecognized by the msvc++ compiler.

MIvanchev: Hey, which VS is this?

MIvanchev:
I've changed the code a bit, maybe you friends can check!

TheForceEngine/TFE_Audio/midiPlayer.cpp
mlauss2: I tried with VS2022, SDK10, v143 (the latest available), msvc++ compiler.

MIvanchev: Interesting... anyhow I refactored the code so it's not used any more.

mlauss2:
gcc says:
```
midiPlayer.cpp:530:51: error: no matching function for call to 'max(<brace-enclosed initializer list>)'
  530 |                                 timeout = std::max({0, timeout});
/usr/lib/gcc/x86_64-pc-linux-gnu/15/include/g++-v15/bits/stl_algobase.h:257:5: note: candidate: 'template<class _Tp> constexpr const _Tp& std::max(const _Tp&, const _Tp&)'
  257 |     max(const _Tp& __a, const _Tp& __b)
      |     ^~~
/usr/lib/gcc/x86_64-pc-linux-gnu/15/include/g++-v15/bits/stl_algobase.h:257:5: note:   candidate expects 2 arguments, 1 provided
/usr/lib/gcc/x86_64-pc-linux-gnu/15/include/g++-v15/bits/stl_algobase.h:303:5: note: candidate: 'template<class _Tp, class _Compare> constexpr const _Tp& std::max(const _Tp&, const _Tp&, _Compare)'
  303 |     max(const _Tp& __a, const _Tp& __b, _Compare __comp)
      |     ^~~
/usr/lib/gcc/x86_64-pc-linux-gnu/15/include/g++-v15/bits/stl_algobase.h:303:5: note:   candidate expects 3 arguments, 1 provided
In file included from /usr/lib/gcc/x86_64-pc-linux-gnu/15/include/g++-v15/algorithm:61,
                 from midiPlayer.cpp:15:
/usr/lib/gcc/x86_64-pc-linux-gnu/15/include/g++-v15/bits/stl_algo.h:5732:5: note: candidate: 'template<class _Tp> constexpr _Tp std::max(initializer_list<_Tp>)'
 5732 |     max(initializer_list<_Tp> __l)
      |     ^~~
/usr/lib/gcc/x86_64-pc-linux-gnu/15/include/g++-v15/bits/stl_algo.h:5732:5: note:   template argument deduction/substitution failed:
midiPlayer.cpp:530:51: note:   deduced conflicting types for parameter '_Tp' ('int' and 'double')
  530 |                                 timeout = std::max({0, timeout});
      |                                           ~~~~~~~~^~~~~~~~~~~~~~
/usr/lib/gcc/x86_64-pc-linux-gnu/15/include/g++-v15/bits/stl_algo.h:5742:5: note: candidate: 'template<class _Tp, class _Compare> constexpr _Tp std::max(initializer_list<_Tp>, _Compare)'
 5742 |     max(initializer_list<_Tp> __l, _Compare __comp)
      |     ^~~
/usr/lib/gcc/x86_64-pc-linux-gnu/15/include/g++-v15/bits/stl_algo.h:5742:5: note:   candidate expects 2 arguments, 1 provided

```

same for next line with std::min().  I think gcc expects a comparison function as 2nd parameter.

Why so complicated though, this here works well too, both clang and gcc are happy with it (no overflow/truncation warnings):
```
                                u64 sleeptime = (s_midiCallback.timeStep - s_midiCallback.accumulator) * 1000;
                                u32 timeout = sleeptime >= SDL_MUTEX_MAXWAIT ?: (u32)sleeptime;
                                SDL_CondWaitTimeout(s_midiCmdCond, s_mutex, timeout);
```

MIvanchev:
I'm not sure if infinities and negative values are undefined behavior in that case, otherwise a good solution unless someone changes the types of timeStep/accumulator. In general, C++ has facilities to make the code a little bit safer but they don't seem to be well supported 🤭

MIvanchev:
OK the deduction should work now, it's hard to test for me RN cuz I'm on a very limited PC... it seems to work with GCC.

mlauss2:
> OK the deduction should work now, it's hard to test for me RN cuz I'm on a very limited PC... it seems to work with GCC.

Yes both clang and gcc are now happy.

mlauss2 (approved):
Looking good, both clang and gcc are happy.

mlauss2:
@luciusDXL please consider to include this PR into the next release.  It makes a world of difference wrt. CPU usage.

jerethk:
I see this has been superseded by #618 so I shall close it.

PR #407, opened 2024-05-17
TFE_Asset: fix copy-paste bug in hd sprite asset cleanup

mlauss2:
This fixes a copy-paste bug which leads to a crash on shutdown when no HD assets are available.
```
Thread 1 "theforceengine" received signal SIGSEGV, Segmentation fault.
TFE_Sprite_Jedi::freePool (pool=pool@entry=POOL_GAME) at TheForceEngine/TFE_Asset/spriteAsset_Jedi.cpp: 717
717                             for (s32 e = 0; e < hdWaxList[i]->entryCount; e++)
```


luciusDXL:
I fixed this in another diff, so closing.

PR #410, opened 2024-06-03
Linux: default linkage to legacy OpenGL (libGL.so) library

### LINKED ISSUE #395: [Linux] Native Linux version launch but black screen
All compiled good
But after launch ./theforceengine only black screen

```
freeartist-devuan@home:~/Downloads/TheDarkForces/TheForceEngine-1.09.540/tfe-build$ ./theforceengine
[Main] The Force Engine 
v1.09.530-13-gd029fe2b

[Paths] Program Path: "/home/freeartist-devuan/Downloads/TheDarkForces/TheForceEngine-1.09.540/tfe-build/"
[Paths] Program Data: "/home/freeartist-devuan/.local/share/TheForceEngine/"
[Paths] User Documents: "/home/freeartist-devuan/.local/share/TheForceEngine/"
[Paths] Source Data: "/home/freeartist-devuan/Downloads/TheDarkForces/DarkForces/"
[Startup] TFE_System::init
[Display] Vertical Sync enabled.
[RenderBackend] OpenGL Device Tier: 1
[Startup] TFE_AudioSystem::init
[Audio] SDLAudio using interface 'pulseaudio'
[Audio] Device 00: Built-in Audio Analog Stereo
[Audio] Starting up audio stream for device '<autoselect>'
[Startup] TFE_MidiPlayer::init
[Startup] TFE_Polygon::init
[Startup] TFE_Image::init
[Startup] TFE_FrontEndUI::init
[MemoryRegion] Allocated new memory block in region 'game' - new size is 1 blocks, total size is '8388608'
[MemoryRegion] Allocated new memory block in region 'level' - new size is 1 blocks, total size is '8388608'
[Progam Flow] The Force Engine Game Loop Started
```

Screenshot_2024-03-13_02-57-51

mlauss2:
Modern CMake (and when I initially wrote the CMakeLists) prefers GLVND, which is a fancy way to dispatch OpenGL calls over multiple different client libraries.  However that does not work with the ancient (<=340.xx) nvidia binary linux drivers, which still assume that there can be only one libGL.so on a linux system.

Even if the GLVND libraries are available on the target system, there will not be any GL output since the client library does not understand the GLVND dispatches, resulting in black screens.

Fix this by explicitly requesting CMake to prefer the legacy OpenGL linkage approach.

Fixes #395 and I suspect a few other crashes as well.


mlauss2:
I'm going to solve this differently, without build-time dynamic linking to libGL.

PR #411, opened 2024-06-04
Fixes for GCC LTO (Link Time Optimizations)

mlauss2:
While building with -flto and gcc on Linux does succeed, the resulting binary crashes early during sound loading due to the Allocator returning nullptrs.  This set of 3 patches removes all warnings emitted by GCC during the LTO run, and the resulting binary actually works fine as well (played a few levels, mods, saved/loaded savegames from both the lto-patched and vanilla binary, no issues whatsoever).

Comparison of binary size (-O3 -march=znver4):
```
   text    data     bss     dec     hex filename
5754171   84401 3355328 9193900  8c49ac theforceengine-gcc15
4317669   48488 3215256 7581413  73aee5 theforceengine-gcc15-lto
```
Size reduction of 1.54MiB or ~17,5%

The thing that LTO was most confused by is apparently the wraparound-to-nullptr design in the TFE_Jedi Allocator,  I changed it a bit to use regular nullptrs.
The rest is namespace cleanups to get rid of C++ ODR violations which only appear with lto enabled,
and nullpointer checks to call allocator_newItem() callsites (which fixes an lto-only allocator pointer corruption encountered with dark tide 1 mod).

Clang LTO also works, but generates a 10% bigger binary (compared to non-lto build).

### DIFF
--- a/TheForceEngine/TFE_Asset/levelObjectsAsset.cpp
+++ b/TheForceEngine/TFE_Asset/levelObjectsAsset.cpp
@@ -17,7 +17,7 @@ namespace TFE_LevelObjects
 
 	bool parseObjects();
 	ObjectClass getObjectClass(const char* name);
-	LogicType getLogicType(const char* name);
+	AssetLogicType getLogicType(const char* name);
 
 	bool load(const char* name)
 	{
@@ -112,7 +112,7 @@ namespace TFE_LevelObjects
 
 			// Logics
 			const u32 logicCount = (u32)obj->logics.size();
-			const Logic* logic = obj->logics.data();
+			const AssetLogic* logic = obj->logics.data();
 			bool isScenery = false;
 			for (u32 l = 0; l < logicCount; l++, logic++)
 			{
@@ -193,7 +193,7 @@ namespace TFE_LevelObjects
 		u32 objectIndex = 0;
 		u32 podIndex=0, sprIndex=0, fmeIndex=0, sndIndex=0;
 		LevelObject* object = nullptr;
-		Logic* logic = nullptr;
+		AssetLogic* logic = nullptr;
 		EnemyGenerator* generator = nullptr;
 		bool sequence = false;
 		while (bufferPos < len)
@@ -453,13 +453,13 @@ namespace TFE_LevelObjects
 		return CLASS_INAVLID;
 	}
 
-	LogicType getLogicType(const char* name)
+	AssetLogicType getLogicType(const char* name)
 	{
 		for (u32 i = 0; i < LOGIC_COUNT; i++)
 		{
 			if (strcasecmp(name, c_logicName[i]) == 0)
 			{
-				return LogicType(i);
+				return AssetLogicType(i);
 			}
 		}
 		assert(0);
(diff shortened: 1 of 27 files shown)


luciusDXL:
Seems useful. This mostly touches the reverse-engineered code (including the Jedi allocator). I went through the changes, but might need to patch and test it for a bit to be sure. :)

mlauss2:
I did build and run test it on windows (together with the other PRs), and it does run fine there as well.
I don't know whether MSVC supports lto though :)

luciusDXL:
> I did build and run test it on windows (together with the other PRs), and it does run fine there as well. I don't know whether MSVC supports lto though :)

I'm not too worried about that, as long as the code compiles and runs correctly (and with the same performance). After some additional testing, I will probably merge it in after the next release (sometime this week).

mlauss2:
I've played all the "kyber krystal 7" levels with the remaster assets and a gcc-15-lto built binary without any issues/crashes.

mlauss2:
I consider this now ready.

luciusDXL:
Ok, thanks. I will test it later today.

luciusDXL (approved):
Thanks for all of the fixes.

luciusDXL:
So it turns out part of the duplication issue was that old files were still in TFE_Asset/ that are not included in the Visual Studio solution which partially duplicated some of the functionality found elsewhere. So I pushed a diff that removes those files, cleans up the CMakeLists in TFE_Assets/ (the file it removed from the sources no longer exists) - and plan to do further cleanup of that code in the near future.

Some of that code existed before the reverse-engineering was done or during the initial phases, which is why it isn't used and needs further cleanup.

mlauss2:
Ah makes sense, yes.  I think vocAsset.cpp is also dead code (but a more complete voc parser than the one in imdigitalsound.cpp).
I think the sound asset system makes sense: decode the voc to a raw wave, maybe even upsample it to 44.1kHz for a common base, and have imuse just play the wave sound. This way you could (as a TFE extension) support higher quality sound assets in mods (and possibly outlaws).

luciusDXL:
> Ah makes sense, yes. I think vocAsset.cpp is also dead code (but a more complete voc parser than the one in imdigitalsound.cpp). I think the sound asset system makes sense: decode the voc to a raw wave, maybe even upsample it to 44.1kHz for a common base, and have imuse just play the wave sound. This way you could (as a TFE extension) support higher quality sound assets in mods (and possibly outlaws).

That was basically the original plan, but I decided to go with the reverse-engineered code as an initial step to make sure it was accurate. With upcoming Outlaws support I have to add wave support anyway, so I plan on implementing HD audio support for Dark Forces/iMuse as well at that point and that can be revived.

mlauss2:
> > Ah makes sense, yes. I think vocAsset.cpp is also dead code (but a more complete voc parser than the one in imdigitalsound.cpp). I think the sound asset system makes sense: decode the voc to a raw wave, maybe even upsample it to 44.1kHz for a common base, and have imuse just play the wave sound. This way you could (as a TFE extension) support higher quality sound assets in mods (and possibly outlaws).
> 
> That was basically the original plan, but I decided to go with the reverse-engineered code as an initial step to make sure it was accurate. With upcoming Outlaws support I have to add wave support anyway, so I plan on implementing HD audio support for Dark Forces/iMuse as well at that point and that can be revived.

Cool .. let me know if I can help here

PR #412, opened 2024-06-05
A11y: call invalidate font atlas indirectly

mlauss2:
Don't call the imgui impl function directly in a11y code, use the helper in TFE_Ui instead.
@kevinfoley I think this is what you initially intended?


mlauss2:
I'm going to fold that into another PR.

PR #413, opened 2024-06-05
Add glFinish() before buffer swap

JakeSmarter:
Just for the sake of completeness (`SDL_GL_SwapWindow()` does not call `glFinish()` for you).
In my understanding, OpenGL commands are *latently asynchronous* by design but please correct if I am wrong. So far, I have not come across any visible negative side effects because of `glFinish()` being absent on SoC (CPU and GPU integrated) systems but could imagine on systems with dedicated GPUs.

### DIFF
--- a/TheForceEngine/TFE_RenderBackend/Win32OpenGL/renderBackend.cpp
+++ b/TheForceEngine/TFE_RenderBackend/Win32OpenGL/renderBackend.cpp
@@ -255,6 +255,7 @@ namespace TFE_RenderBackend
 
 		TFE_ZONE_BEGIN(swapGpu, "GPU Swap Buffers");
 		// Update the window.
+		glFinish();
 		SDL_GL_SwapWindow((SDL_Window*)m_window);
 		TFE_ZONE_END(swapGpu);
 


mlauss2:
From Khronos:
`glFinish does not return until the effects of all previously called GL commands are complete. Such effects include all changes to GL state, all changes to connection state, and all changes to the frame buffer contents.`

It blocks the CPU until GPU has finished rendering. That time could be used to prepare the next frame.

`glFinish requires a round trip to the server. `
also a perf killer.

mlauss2:
> > It blocks the CPU until GPU has finished rendering. That time could be used to prepare the next frame.
> 
> That’s right. However, how do know when rendering has finished before you swap buffers? Note that 

I think SDL (at least glXSwapInterval()) will do a glFlush() when you call SDL_SwapBuffers(), which works like glFinish() but the app can immediately start submitting new GL commands (which only take effect after the buffer swap has completed).

> > `glFinish requires a round trip to the server.`
> > also a perf killer.
> 
> It is a synchronization tool, not a GL command to “kill perf”. 😉

I think even if there is nothing to do/render, it will take comparatively large amount of time.

PR #414, opened 2024-06-07
[RFC] get rid of GLEW requirement

mlauss2:
Get rid of GLEW in TFE completely.
- Update Dear ImGui to 1.90.8 (release 06-JUN-2024), which dropped GLEW in favour of pregenerated headers, plus fixups for changed interfaces
- - added a Bugfix to ImGui to get A11y runtime font changes working again (seems to be a genuine bug in imgui).
- Drop GLEW usage from the codebase.  GL Feature detection is done with SDL GL Extension detection instead.

On Linux, this has the added benefit that TFE can now run on any windowing platform that the underlying system SDL library supports. E.g. I can now play TFE directly from the console (i.e. gbm/EGL) without requiring any X-Server or Wayland compositor running: (Example from an old 32bit atom n270 with nvidia ION):
image

There's still some bits missing to make this work completely (mainly, replacing PortableFileDialogs with something ImGUI-based).


mlauss2:
I stared at this a bit more and am going to implement this differently.

PR #416, opened 2024-06-23
Set Application name SDL Hint

mlauss2:
Set the "Application Name" SDL Hint.  This fixes a cosmetic issue on at least Linux wrt. to the audio stream name when pipewire/pulseaudio are used as sound servers (before/after):
image


mlauss2:
Encased it in an ifdef since the windows sdl2 in the TFE tree doesn't have this hint defined.

luciusDXL (approved):
Thanks.

PR #419, opened 2024-07-08
More bugfixes/improvements

mlauss2:
More bug/warningfixes and updates, also to get it building on Apple/Clang toolchains.  based on feedback in #90 

- QUAD_MAX is a reserved symbol in xnu's "limits.h" file
- clang doesn't like passing std::string to logWrite()
- disable clang's excessive warnings about unhandled enum values in switch() statements
- fix lots of clang warnings about assignments inside while() without extra brackets.
- adjust the min/max/clamp(size_t) functions to u64, since size_t == u32 on 32bit targets
- LINE_MAX is another reserved symbol on xnu/apple.
- change the size_t in memoryRegion, which is written out to savegames, to an u64, to keep apple toolchain happy and the savegame format independent of the target architecture size.
- rename the "\<autoselect\>" audio device to "System Default", and change the System Midi "(Disabled)" device to explain why it's disabled (By choice, no synthesizers, or RtMidi error).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,12 +62,13 @@ if(ENABLE_TFE)
 	if(UNIX)
 		find_package(PkgConfig REQUIRED)
 		find_package(SDL2 2.0.20 REQUIRED)
-		find_package(SDL2_Image REQUIRED)
+		pkg_check_modules(SDL2_IMAGE REQUIRED SDL2_image)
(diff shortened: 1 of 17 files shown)


jodaco:
Definitely get much further with this PR on Mac, sent a minor PR to your original branch for CMakeLists to at least get the build going on Mac

luciusDXL (approved):
Looks good, thanks for all of the improvements.

PR #418, opened 2024-06-27
Drop GLEW in favour of built-in GLAD

mlauss2:
Drop GLEW in favor of GLAD-generated built-in loader.
- This drops an extra external library dependency (glew)
- On Linux no system OpenGL libraries/headers are required to build TFE any more
- All the platform specific stuff (windowing, library loading, symbol
  resolution) is now fully handled by SDL2, no more wayland workarounds necessary (i.e. #386 )
    
The GLAD loader is GL4.6, while TFE's minimum is still GL3.3;
the Anisotropy stuff, which is core GL4.5, has been wrapped around
availability of EXT_texture_filter_anisotropic (which most GL2
drivers support).

This PR also updates ImGUI to version 1.90.9 because after the current in-tree 1.72 it also dropped GLEW.

This should also fix #395  which was caused by TFE explicitly linking to an OpenGL library, which, after this PR, is no longer necessary.

After applying this PR, the dependencies change to:
```diff
-Shared library: [libOpenGL.so.0]
-Shared library: [libGLX.so.0]
-Shared library: [libGLU.so.1]
-Shared library: [libGLEW.so.2.2]
-Shared library: [libGL.so.1]
-Shared library: [libX11.so.6]
 Shared library: [libSDL2-2.0.so.0]
 Shared library: [libSDL2_image-2.0.so.0]
 Shared library: [librtmidi.so.7]
```
i.e. no explicit dependency on OpenGL and windowing system, just SDL2.

Another nice sideeffect at least on Linux is that "zink" (OpenGL over Vulkan) also works now:
image
Performance on my radeon is about 66% of the direct opengl path.

PR #420, opened 2024-07-13
ImGuiFileDialog to replace portable-file-dialogs

mlauss2:
So here's a first stab at implementing ImGuiFileDialog (https://github.com/aiekick/ImGuiFileDialog) to generate the file/directory selection dialogs.
Applies on top of PR #418 (which updates ImGui to 1.90.9) / this PR comes with #418 included since I used it as development base, and the Dialog needs at least ImGui-1.90.5

The Dialog uses std::filesystem, which means a compiler with c++17 support is required.

At the moment it looks like this:
tfe-imguidialog

It can be resized, I've for now chosen minimum size of 640x480px (image above), any smaller and
interacting with it becomes tedious, I found myself first resizing it every time it opened.
I'll look into having Icons instead of [File]/[Dir] and maybe different colors for files and dirs.

Tested on Windows and Linux, including the Editor (but that might still have bugs).
While this is no big deal for windows, on linux this again reduces external dependencies (external dialog app), and tfe+editor fully work without any desktop environment.


mlauss2:
Added Icons and a bit of coloring:
tfe_filedlg2

luciusDXL:
I remember trying this in the past and not liking it as much, but the extra dependencies on Linux is a good point and it is more consistent in terms of UI.

mlauss2:
yeah it's not the prettiest interface but outside of the editor you're probably going to use it only once, on initial setup.  My main motivation was to get tfe running on my kodi box which doesn't have any windowing system (i.e. runs directly on the DRM framebuffer) and now all including the editor is usable.

PR #421, opened 2024-07-15
Update Dear ImGUI to 1.90.9

mlauss2:
- Update ImGUI to 1.90.9
- export the custom TFE widgets to a separate source file to make imgui updates simpler
- Fix up the tree to accomodate all ImGUI interface changes

Main motivation is 
- sometime after 1.72 in the tfe tree, imgui dropped GLEW in favour of a built-in GL loader
- I'd like to get rid of GLEW use in TFE
- I'd like to replace portable-file-dialogs with ImGuiFileDialog (which needs a newer ImGui than what is in-tree)

Settings pages and Editor (at least the parts of the editor that I'm smart enough to use :) ) all work fine.

This supersedes #418   and I'll add another PR if @luciusDXL is fine with this one.


luciusDXL:
Looks good, I will patch it in after the next release.

mlauss2:
I went over the patch once again, the Editor, esp. the Modal Popups should be working again, and rebased it on top of master.

This version of ImGui has problems(?) with Modal Popups that you also want to have a close button on.
The popup isn't rendered since it's not the topmost in the popup stack and the variable to track the close button is immediately set to false, hiding the popup in the next frame (where it finally IS the topmost).

So I dropped this parameter from every Modal Popup and instead added a "Close" Button at the bottom.

luciusDXL:
> I went over the patch once again, the Editor, esp. the Modal Popups should be working again, and rebased it on top of master.
> 
> This version of ImGui has problems(?) with Modal Popups that you also want to have a close button on. The popup isn't rendered since it's not the topmost in the popup stack and the variable to track the close button is immediately set to false, hiding the popup in the next frame (where it finally IS the topmost).
> 
> So I dropped this parameter from every Modal Popup and instead added a "Close" Button at the bottom.

Annoying but good to know.

luciusDXL (approved):

luciusDXL:
The PR was merged in. I have been fixing up Editor issues before pulling in the File dialog change as well.

mlauss2:
Thanks for fixing the Editor!

luciusDXL:
> Thanks for fixing the Editor!

The Editor is already pretty complex, so I'm not surprised there were issues. I also fixed some non-imGUI related issues I found as well (my own bugs) so it was a good idea to go through all of the UI anyway.

PR #422, opened 2024-07-16
IMuse: handle type 2 VOC  chunks

mlauss2:
In Dark Tide 1, there's the MCHN3.VOC file which contains a voc chunk with id 2, which causes TFE to spam the log with an unending stream of:

```
[Error : iMuse] ERR: Illegal chunk in sound    2561ba3c54df0...
[Error : iMuse] ERR: Illegal chunk in sound    3561ba3cb5658...
[Error : iMuse] ERR: Illegal chunk in sound    5561ba3cb5658...
```

Fix this by handling type 2 chunks.

Also clarify the debug output a bit and stop spamming the log on known but unhandled chunks (3 and 5).

In DT1, there's also the "BUBBLS1.VOC" file which has only 1 chunk but its size is too small, doesn't cover the entire file. This will cause a one-time log entry
```
[Error : iMuse] ERR: Illegal chunk 170 in sound 657844099906576...
```

### DIFF
--- a/TheForceEngine/TFE_Jedi/IMuse/imDigitalSound.cpp
+++ b/TheForceEngine/TFE_Jedi/IMuse/imDigitalSound.cpp
@@ -486,9 +486,12 @@ namespace TFE_Jedi
 			{
 				return imFail;
 			}
-			else if (id == 1)	// found the next useful chunk.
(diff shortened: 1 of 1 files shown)


luciusDXL (approved):

luciusDXL:
Looks good, thanks.

PR #423, opened 2024-07-17
CMake: automatically make system midi optional

mlauss2:
- Flip the system midi CMake configure option to "ENABLE_" like all others and enable it by default.
- Only If ENABLE_SYSMIDI evaluates to true try to find RtMidi.
- If RtMidi is unusable, disable the system midi output module.

This makes RtMidi an optional feature in the linux/unix build.  If it's available at build time, it will be included, but no longer fail if it isn't, since the built in OPL3 and SF2 renderers are already very good.


luciusDXL (approved):
Looks good to me.

PR #424, opened 2024-07-19
Replace GLEW with built-in GLAD

mlauss2:
Add a GLAD-generated opengl loader to the OpenGL renderbackend, to remove the dependency on external GLEW.

This should not be noticeable on Windows, but is beneficial on Linux for several reasons:
- at build time, no system opengl libraries and headers are required.
- TFE Linux does no longer have any built-in OpenGL or windowing system dependecy, just SDL2.
- all the ugly GL and Windowing stuff is left to SDL.


luciusDXL:
Looks pretty good, I will test it out tomorrow and merge if everything looks ok.

luciusDXL (approved):

luciusDXL:
I patched this in locally and it works great on Windows, so I merged it in. Having one less external dependency (glew) is not something I'm going to complain about. :)

PR #435, opened 2024-07-27
More little fixes

### LINKED ISSUE #430: Linux (Arch Linux) x86_64 - Segfault when launching game
Hi, I can compile and run version v1.09.540 with no issues, however when compiling either v1.10.000 or the current master branch I get a segfault error when running the executable. I am using an old Dell Latitude with Intel graphics so I am enquiring if the requirements for OpenGL or Vulkan (If used) have changed?

The commands I use to compile theforceengine are:

cmake . \
        -B"build" \
        -G"Ninja" \
        -DCMAKE_BUILD_RPATH_USE_ORIGIN="ON" \
        -DCMAKE_BUILD_TYPE="Release" \
        -DCMAKE_C_COMPILER="clang" \
        -DCMAKE_CXX_COMPILER="clang++" \
        -DCMAKE_EXE_LINKER_FLAGS_INIT="-fuse-ld=lld" \
        -DCMAKE_INSTALL_PREFIX="/opt/archypie/ports/forceengine" \
        -DCMAKE_MODULE_LINKER_FLAGS_INIT="-fuse-ld=lld" \
        -DCMAKE_SHARED_LINKER_FLAGS_INIT="-fuse-ld=lld" \
        -Wno-dev
ninja -C build

If no hardware requirements have changed please let me know and I will provide a backtrace. Thanks

### LINKED ISSUE #433: [🐛] Custom resolution resets to matching
Linux Flathub v1.10.000-28-ge19bcde1+

Whenever you return to the graphics settings menu, any custom resolution is reset to matching. However, the custom resolution is saved and loaded whenever you launch a “The Force Engine” instance, as long as you do not go into the graphics settings menu.

Furthermore, setting a custom resolution width basically does nothing for you, that is, it does not automatically change the height with the correct aspect ratio applied. However, setting the height does automatically change the width with the correct aspect ratio applied. But, the width UI field does not visibly change its value. :face_with_spiral_eyes:

mlauss2:
More little things encountered so far:
*  move the sdl2main.lib #pragma under the _WIN32 ifdef, because the lld linker on not-windows adheres to it
* Abort DF launch early if not all GOBs can be found, to avoid a crash later. Found while playing with flatpak.
* address a new -Wformat-security warning in the Editor code
* constify the pointers in the game source data header even more to fix a broken lld, fixes #430, although in my opinion this is a problem with the OPs linker, not TFE, but the fix is harmless.

### DIFF
--- a/TheForceEngine/TFE_DarkForces/darkForcesMain.cpp
+++ b/TheForceEngine/TFE_DarkForces/darkForcesMain.cpp
@@ -230,7 +230,7 @@ namespace TFE_DarkForces
 	void setInitialLevel(const char* levelName);
 	s32  loadLocalMessages();
 	void buildSearchPaths();
-	void openGobFiles();
+	bool openGobFiles();
 	void gameStartup();
 	void loadAgentAndLevelData();
 	void startNextMode();
@@ -295,7 +295,8 @@ namespace TFE_DarkForces
 		buildSearchPaths();
 		processCommandLineArgs(argCount, argv, startLevel);
 		loadLocalMessages();
-		openGobFiles();
+		if (!openGobFiles())
+			return false;
 
 		// Sound is initialized before the task system.
 		sound_open(s_gameRegion);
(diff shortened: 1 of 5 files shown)


luciusDXL (approved):
Thanks for the fixes.

PR #438, opened 2024-08-13
Windows ARM64 build target

mlauss2:
Add a windows-arm64 target to the VC Solution file
- Delete the SDL2.dll and SDL2_image.dll from main folder
- Add arm64 versions of SDL2 libs+dlls
- Add ARM64 Release and Debug targets to solution
- Add a post-link step to copy the target architecture's
  SDL2.dll and SDL2_image.dll to main folder
- Fix up AngelScript MASM-ARM64 source file to actually build.


mlauss2:
Run-tested on a Qualcomm X1E device.  It's playable but buggy/wonky:
Probably TFE issue:
- adding an agent works but agent name isn't displayed in the list

probably HW/driver issues:
- keyboard input is slow/missing keys sometimes (the EFI shell also misses keystrokes, but windows itself is fine)
- imgui during game pause is really slow (it takes 2-3 seconds until it responds to interactions)
- shader compiler is sloooow, clicking "start" takes ~6 seconds until it actually starts.

So it works, but can't be recommended.  Even the 32-bit linux version runs better on my 2009 nvidia ion device.

PR #445, opened 2024-09-08
[RFC] PhysFS integration

mlauss2:
RFC, this rather large change aims to implement PhysFS (https://icculus.org/physfs/) support in TFE.
Physfs is used to create virtual file system tree where all app- and game data are easily accessible.
The main benefits are
- the TFE binary can be shipped with all the support data (Fonts, Soundfonts, UI Images, Texts, ..)
  packed into the binary itself.
- DF and OL can be packed into a single ZIP file and used as source.

Additional Soundfonts, captions, mods, .. just need to be placed in the appropriate .local/share/TheForceEngine/ subdir and are automatically picked up as _additional_ resources to the built-in ones
(i.e. I deliberately chose to not allow to override the shipping data [shaders, ...]).

To get DF working, all you need is a ZIP of the DF directory (optionally place enhanced.gob into it as well),
tfe binary and an empty "settings.ini" to use portable mode (don't compress the df source zip too much, as e.g. zip compression level 9 raises level loading time from <1s to almost 6s!)
tfezip-01

https://github.com/user-attachments/assets/5b96087a-3008-452b-a0f4-0e12cee33d63

At the moment main game works, including saving + loading.
I haven't touched the modloader or editor yet.
Not tested on windows.

Comments welcome!


luciusDXL:
Is it possible to continue using the existing setup with this change? I'm not sure I like the idea of packing all of the data into the executable, I like having it visible rather than hiding it (at least on Windows).

mlauss2:
Yes it is possible.  The code tries to first open the executable to look for its support data, checks for a sentinel file, then repeats the same thing with the current directory (i.e where binary was started from).

Issue #449, opened 2024-09-20
[Feature] Add Mod Overrides Handling via MOD_CONF.txt


ifeldshteyn:
At the moment the MOD_CONF is used to override game mechanics but there is no way to handle player settings on a per-mission status. This was prompted by being able to start a mission only with your bryar pistol like the Secret Base.

Issue #446, opened 2024-09-17
[Feature] Refresh the Mod Loader in TFE


ifeldshteyn:
At the moment, for caching reasons, once TFE loads the mods you add or remove any more. You have to restart the whole program. This need a refresh button.

PR #450, opened 2024-09-22
Adding Mod Refresh option and expanding MOD_CONF overrides

ifeldshteyn:
This PR handles three issues.

1. Addition of a mod refresh button from the mod menu for #446 
2. Initial commit for level MOD_CONF overrides. These overrides allow you to specify what weapons and ammo and health you start with. You can either start the mission wearing your gas mask. Details are here ---> https://df-21.net/wiki/?title=TFE_Mod_Overrides This is for #449 
3. Addition of mouse panning in the PDA #453 

Adding a simple feature. Timestamps in the logs for linux/windows.


TheForceEngine/TFE_FrontEndUI/modLoader.cpp
```diff
@@ -90,7 +90,8 @@ namespace TFE_FrontEndUI
 
 	void modLoader_read()
 	{
-		if (s_modsRead) { return; }
+		// Reuse the cached mods unless no mods have been read yet.
+		if (s_modsRead and s_mods.size() > 0) { return; }
```
luciusDXL, line 94: I'm surprised this works, it should be '&&' instead of 'and'.

ifeldshteyn: It 100% works. Apparently they are alternatives https://stackoverflow.com/questions/47174920/difference-between-and-and-operators . I will switch to && for consistency (I am coming from a python background).

jerethk: C# has implemented "and" and "or"  as part of its pattern matching 

Instead of `if (a == b || a == c)`
you can now do `if (a is b or c)`

I still find it disconcerting...

ifeldshteyn: Oh that's neat - I love it.

mlauss2 (commented):
I suggest you move the functions out of a11y and alongside _strlwr() and co at TFE_System/system.cpp/h

ifeldshteyn:
> I suggest you move the functions out of a11y and alongside _strlwr() and co at TFE_System/system.cpp/h

100% agree. I was wondering why they were in A11y in the first place. I will let Lucius make the change

PR #451, opened 2024-10-02
Custom logic proposal

jerethk:
The aim is to give people a simple way to define custom enemy/AI logics in an external JSON file

- An actor setup method that creates an ActorDispatch logic from custom values
- A data structure that holds custom logic values (VOC file names, hitpoints, projectile, drop item, attack range, etc.)
- Code which parses JSON and loads the custom logic data into settings
- Code which will search for a matching custom logic from a list and then call the setup method if found
- Amendments have been made to support custom logic use with generators
- Amendments to the attack function (in `Actor.cpp`) allowing mortars fired by AI to behave like thermal detonators (fired upwards in an arc)
- Custom logics will also be loaded from inside mod ZIPs


TheForceEngine/TFE_DarkForces/Actor/actor.cpp
```diff
@@ -1012,23 +1013,40 @@ namespace TFE_DarkForces
 				SecObject* projObj = proj->logic.obj;
 				projObj->yaw = obj->yaw;
 				
-				// Handle x and z fire offset.
-				if (attackMod->fireOffset.x | attackMod->fireOffset.z)
+				if (attackMod->projType == PROJ_THERMAL_DET || attackMod->projType == PROJ_MORTAR)
```
luciusDXL, line 1016: Does this change vanilla behavior? If so it will need to be made optional.

TheForceEngine/TFE_Settings/dfLogics.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+namespace TFE_Settings
+{
+	///////////////////////////
+	/// Forward declarations 
+	///////////////////////////
+	bool tryAssignProperty(cJSON* data, CustomActorLogic& customLogic);
+
+	
+	void parseLogicData(char* data, std::vector<CustomActorLogic> &actorLogics)
```
luciusDXL, line 17: Super nitpicky, but in the code I usually pair the * or & with the type for pointers/references, aka:
char* data, std::vector<...>& actorLogics.

TheForceEngine/TFE_Settings/settings.cpp
```diff
@@ -1483,4 +1503,37 @@ namespace TFE_Settings
...
 	}
+
+	void loadCustomLogics()
+	{
+		vector<string> jsons;
+		TFE_Paths::getAllFilesFromSearchPaths("logics", "json", jsons);
+
+		for (u32 i = 0; i < jsons.size(); i++)
+		{
+			TFE_System::logWrite(LOG_MSG, "LOGICS", "Parsing logic JSON");
```
luciusDXL, line 1514: Do we need to write log spam for each logic? It might be better to write at the top of the loop. You can write the number parsed at the end if desired.

luciusDXL (approved):

TheForceEngine/TFE_DarkForces/Actor/actor.cpp
jerethk: It does not change vanilla behaviour, because none of the vanilla characters have thermal detonators as their primary attack (`case STATE_FIRE1:`)

To  clarify the change I've made here:
The original code is found on lines 1033-1049, exactly as it was before, but now within the `else` code block. So it now follows the same logic as occurs with `STATE_FIRE2`

`if (arcing projectile)
{
     aimArcing
}
else
{
    aimAtTarget
}`

Vanilla enemies will all fall through to the `else` block (i.e. the original code) because none of them use an arcing projectile as their `STATE_FIRE1`. 

Hope I've explained that adequately

TheForceEngine/TFE_Settings/dfLogics.cpp
jerethk: Sure, will amend this

jerethk:
I have squashed all my existing commits into a single commit, rebased on master, and fixed up the merge conflicts.

TheForceEngine/TFE_DarkForces/Actor/actor.cpp
luciusDXL: Ok, thanks for the explanation - it makes sense now.

TheForceEngine/TFE_Settings/settings.cpp
luciusDXL: Once you move this out of the loop, I will squash and push the changes. :)

Thanks for adding this feature!

jerethk:
@luciusDXL this is ready for another review
I've made the amendments you requested, plus a few other minor tweaks 
Also decided to move all the new custom logic stuff out of TFE_Settings and into a new directory & namespace called TFE_ExternalData

I will be using TFE_ExternalData when I externalise the projectile & weapon data (the next project!!)

I think it would be better to have all this related code live together rather than piling it into Settings.

TheForceEngine/TFE_DarkForces/generator.cpp
```diff
@@ -211,12 +230,13 @@ namespace TFE_DarkForces
...
 	}
 
-	Logic* obj_createGenerator(SecObject* obj, LogicSetupFunc* setupFunc, KEYWORD genType)
+	Logic* obj_createGenerator(SecObject* obj, LogicSetupFunc* setupFunc, KEYWORD genType, const char* logicName)
 	{
 		Generator* generator = (Generator*)level_alloc(sizeof(Generator));
 		memset(generator, 0, sizeof(Generator));
 
 		generator->type   = genType;
+		generator->logicName = string(logicName);
```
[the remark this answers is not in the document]
mlauss2: I think you should change this member from "string logicName" to "char logicName[16]" and do "strncpy(generator->logicName, logicName, 15);" (or whatever the max length of a logic name is).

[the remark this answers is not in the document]
mlauss2: Add something like this on top of this patchset, it survived a few hours of playtesting.
Don't use std::string since the only place you use this member you do ".c_str()" on it anyway.

```diff --git a/TheForceEngine/TFE_DarkForces/generator.cpp b/TheForceEngine/TFE_DarkForces/generator.cpp
index e5b9878b..83b47218 100644
--- a/TheForceEngine/TFE_DarkForces/generator.cpp
+++ b/TheForceEngine/TFE_DarkForces/generator.cpp
@@ -40,8 +40,7 @@ namespace TFE_DarkForces
                Tick       wanderTime;
                Wax*       wax;
                JBool      active;
-
-               string          logicName;              // JK: added to store a custom logic name
+               char       logicName[32];               // JK: added to store a custom logic name
        };
 
        void generatorTaskFunc(MessageType msg)
@@ -118,7 +117,7 @@ namespace TFE_DarkForces
 
                                        // Search the externally defined logics for a match
                                        TFE_ExternalData::CustomActorLogic* customLogic;
-                                       customLogic = tryFindCustomActorLogic(gen->logicName.c_str());
+                                       customLogic = tryFindCustomActorLogic(gen->logicName);
                                        if (customLogic && gameSettings->df_jsonAiLogics)
                                        {
                                                obj_setCustomActorLogic(spawn, customLogic);
@@ -236,7 +235,8 @@ namespace TFE_DarkForces
                memset(generator, 0, sizeof(Generator));
 
                generator->type   = genType;
-               generator->logicName = string(logicName);
+               memset(generator->logicName, 0, 32);
+               strncpy(generator->logicName, logicName, 31);
                generator->active = 1;
                generator->delay  = 0;
 

```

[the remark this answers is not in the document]
jerethk: @mlauss2 thanks for that recommendation -- I've got a couple other PRs waiting review, I'll look at this as soon as those have gone through.

PR #454, opened 2024-10-16
GIF capture feedback messages

kevinfoley:
Display UI to let the user know when GIF recording starts and ends.

This PR also includes two corresponding settings in the System settings menu.

<s>When we press Alt-F2 to start recording...
- If "Show countdown before recording starts" is enabled, a message appears in the upper-left corner, counting down from 3 over 2 seconds. During the countdown, the user can press Alt-F2 again to cancel recording. Otherwise, when the countdown finishes, recording starts.
- If "Show countdown before recording starts" is disabled, the word "REC" flashes in the upper-left corner for approximately 1/10th of a second, and then recording starts.</s>

**Edit:**
Now there are two keybindings:

- If we press the "GIF Recording" binding, a message appears in the upper-left corner, counting down from 3 over 2 seconds. During the countdown, the user can press the binding again to cancel recording. Otherwise, when the countdown finishes, recording starts.
- If we press the "Instant GIF Record" binding, the word "REC" flashes in the upper-left corner for approximately 1/10th of a second, and then recording starts.</s>

When we press either keybinding again to stop recording...
- If "Show file path when recording ends" is enabled, a message appears in the upper-left corner for 4 seconds, displaying the path to the saved file.
- If "Show file path when recording ends" is disabled, a message appears in the upper-left corner for 4 seconds, displaying the message "Recording saved".

Additionally, we can pass an argument to the recording function to skip the countdown even if the user has the countdown enabled. This feature is currently unused, but will allow us to have a separate key binding for "start recording without countdown".


TheForceEngine/TFE_RenderBackend/Win32OpenGL/screenCapture.cpp
```diff
@@ -217,24 +218,145 @@ void ScreenCapture::captureFrame(const char* outputPath)
...
 {
-	m_recordingStarted = true;
+	m_capturePath = string(path);
+	if (!skipCountdown && TFE_Settings::getSystemSettings()->showGifRecordingCountdown)
+	{
+		startCountdown();
+	}
+	else
+	{
+		//startCapture();
```
luciusDXL, line 230: No need to keep this comment around.

luciusDXL (approved):

luciusDXL:
Other then the errant comment, this looks good. Unfortunately there is a conflict now due to other PRs that got merged, so if you fix that up I will merge this.

luciusDXL (approved):

PR #457, opened 2024-10-19
Mod conf overrides

ifeldshteyn:
Implementing mod overrides

### DIFF
--- a/TheForceEngine/TFE_DarkForces/mission.cpp
+++ b/TheForceEngine/TFE_DarkForces/mission.cpp
@@ -443,7 +443,9 @@ namespace TFE_DarkForces
 						reticle_enable(true);
 					}
 					s_flatLighting = JFALSE;
-					s_nightvisionActive = JFALSE;
+					// Note: I am not sure why this is there but it overrides all player settings
+					// By default the player load disables night vision so it won't carry over from previous maps.
+					// s_nightvisionActive = JFALSE;
 				}
 			}
 			else // Loading from save.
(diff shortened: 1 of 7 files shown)


TheForceEngine/TFE_DarkForces/mission.cpp
```diff
@@ -443,7 +443,9 @@ namespace TFE_DarkForces
 						reticle_enable(true);
 					}
 					s_flatLighting = JFALSE;
-					s_nightvisionActive = JFALSE;
+					// Note: I am not sure why this is there but it overrides all player settings
```
luciusDXL, line 446: Vanilla often set things to the same setting in multiple places. If it is not needed, it can be removed.

TheForceEngine/TFE_DarkForces/player.cpp
```diff
@@ -447,6 +447,244 @@ namespace TFE_DarkForces
...
+			s_lifeCount = pickup_addToValue(0, intMap["lives"], 9);
+		}
+		if (intMap.find("battery") != intMap.end())
+		{
+			fixed16_16 batteryPower = FIXED(2);
+			int batteryPowerPercent = pickup_addToValue(0, intMap["battery"], 100);
+			s_batteryPower = (batteryPower * batteryPowerPercent) / 100;
+		}
+
+		// Note - this doesn't check if the weapon is is in the inventory!
```
luciusDXL, line 504: Nickpick: is is

luciusDXL (approved):

TheForceEngine/TFE_DarkForces/player.cpp
ifeldshteyn: "is is" is no more.

PR #458, opened 2024-10-19
Adding open folder button for mods.

ifeldshteyn:
Add open folder buttons in the mod section.

### DIFF
--- a/TheForceEngine/TFE_FrontEndUI/modLoader.cpp
+++ b/TheForceEngine/TFE_FrontEndUI/modLoader.cpp
@@ -25,6 +25,10 @@
 #include <map>
 #include <algorithm>
 
+#ifdef _WIN32
+	#include <windows.h>
+#endif
+
 using namespace TFE_Input;
 
 namespace TFE_FrontEndUI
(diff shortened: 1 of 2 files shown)


TheForceEngine/TFE_FrontEndUI/modLoader.cpp
```diff
@@ -90,7 +95,8 @@ namespace TFE_FrontEndUI
 
 	void modLoader_read()
 	{
-		if (s_modsRead) { return; }
+		// Reuse the cached mods unless no mods have been read yet.
+		if (s_modsRead and s_mods.size() > 0) { return; }
```
luciusDXL, line 99: This should be && instead of and

TheForceEngine/TFE_FrontEndUI/modLoader.cpp
```diff
@@ -444,6 +451,21 @@ namespace TFE_FrontEndUI
...
 		}
+
+		ImGui::SameLine(510.0f * uiScale);
+		if (ImGui::Button("Refresh Mod Listing"))
+		{
+			s_modsRead = false;
+			modLoader_read();
+		}
+
+		#ifdef _WIN32
```
luciusDXL, line 462: use osShellExecute() from system.h, that way you don't need to include windows.h here.

luciusDXL (changes requested):

TheForceEngine/TFE_FrontEndUI/modLoader.cpp
ifeldshteyn: Switched to osShellExecute

luciusDXL (approved):

PR #459, opened 2024-10-19
Adding Mouse panning to the PDA

ifeldshteyn:
Add mouse panning to the PDA

### DIFF
--- a/TheForceEngine/TFE_DarkForces/GameUI/pda.cpp
+++ b/TheForceEngine/TFE_DarkForces/GameUI/pda.cpp
@@ -356,7 +356,7 @@ namespace TFE_DarkForces
 		if (TFE_Input::mouseDown(MBUTTON_LEFT))
 		{
 			s32 mouseX, mouseZ;
-			SDL_GetMouseState(&mouseX, &mouseZ);
+			TFE_Input::getMousePos(&mouseX, &mouseZ);
 
 			// Calculate deltas
 			fixed16_16 deltaX = (mouseX - mouseOldPosX) << 16;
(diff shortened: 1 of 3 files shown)


TheForceEngine/TFE_DarkForces/GameUI/pda.cpp
```diff
@@ -350,6 +352,32 @@ namespace TFE_DarkForces
 		// Ensure that mouse movements are not lost between updates.
 		pda_handleMouseAccum(&wdx, &wdy);
 
+		// Handle Mouse Panning
+		if (TFE_Input::mouseDown(MBUTTON_LEFT))
+		{
+			s32 mouseX, mouseZ;
+			SDL_GetMouseState(&mouseX, &mouseZ);
```
luciusDXL, line 359: TFE_Input has a function to get the mouse position, so you don't need to include SDL here.

luciusDXL (changes requested):

TheForceEngine/TFE_DarkForces/GameUI/pda.cpp
ifeldshteyn: This is implemented.

TheForceEngine/TFE_DarkForces/GameUI/pda.cpp
```diff
@@ -1,5 +1,5 @@
 #include <cstring>
-
+#include <SDL.h>
```
luciusDXL, line 2: You should no longer need this include.

luciusDXL (changes requested):

TheForceEngine/TFE_DarkForces/GameUI/pda.cpp
ifeldshteyn: SDL is now removed in [bda14bc](https://github.com/luciusDXL/TheForceEngine/pull/459/commits/bda14bc28b486a53c034c4c8f425474dc863a6d1)

luciusDXL (approved):

PR #464, opened 2024-10-30
Mod conf overrides - VERSION UPDATE

ifeldshteyn:
Starting version count with 1 to make it easier on modders.


ifeldshteyn:
will redo it

PR #466, opened 2024-11-07
Adding rotate 180 option for purple.

ifeldshteyn:
New button to rotate the player 180 degrees ? Will this be possible to just do via modding?


luciusDXL:
I'm going to hold off on this for now, I'm not sure if things like this are worth adding until we see what is possible through scripting.

PR #468, opened 2024-11-12
Fix VUE camera

jerethk:
The (previously unused) camera transform did not calculate pitch, and did not copy pitch to the transformed object. 
It was just set at 0

### DIFF
--- a/TheForceEngine/TFE_DarkForces/vueLogic.cpp
+++ b/TheForceEngine/TFE_DarkForces/vueLogic.cpp
@@ -213,6 +213,12 @@ namespace TFE_DarkForces
 		size_t bufferPos = 0;
 		if (!strcasecmp(transformName, "camera"))
 		{
+			// TFE: allocate a VFRAME_FIRST for camera transforms, same as with ordinary transforms
+			VueFrame* frame = (VueFrame*)allocator_newItem(vueList);
+			if (!frame)
+				return;
+			frame->flags = VFRAME_FIRST;
+
 			while (1)
 			{
 				const char* line = parser->readLine(bufferPos);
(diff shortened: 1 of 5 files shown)


TheForceEngine/TFE_DarkForces/vueLogic.cpp
```diff
@@ -586,6 +597,8 @@ namespace TFE_DarkForces
 
 							local(obj)->posWS = local(frame)->offset;
 							local(obj)->yaw = local(frame)->yaw;
+							local(obj)->pitch = local(frame)->pitch;	// TFE: copy pitch and roll to object (not done in vanilla)
```
jerethk, line 600: @luciusDXL can you please give your opinion about this
In GPU I am happy for the camera/Eye object pitch to ignore pitch limit -- I think that the camera should be able to pitch to 90 degrees even if the player's pitch limit is less.

However, what would the implications be in software renderer, and is there something I should do about that? thanks

luciusDXL: Good question, I think about 60 degrees or so is roughly the useful max for the software renderer. So probably start with something like that initially and maybe add an option to extend it for the GPU renderer later.

jerethk: The approach I have taken is to apply the clamping (to 60 degrees) in jediRenderer.cpp
I think that this is the appropriate place to do it.

An object moved by VUE (which may or may not be the camera) should be able to have any pitch; we apply clamping when it comes to rendering the view, because the motivation is to avoid a heavily distorted view.

luciusDXL (approved):

PR #470, opened 2024-11-19
Custom logics fix

jerethk:
@mlauss2 discovered that the existing code crashes with gcc-14/15
I've replaced std::string with char[] as per his recommendation

note- the source variable (s_objSeqArg2) is has  a length of 256


mlauss2:
Works for me(tm) -- tested with gcc14/15/clang19 on linux x64 and arm64.

luciusDXL (approved):

luciusDXL:
Thanks for the help @mlauss2. :)

PR #476, opened 2024-12-04
Fix custom logics serialisation

jerethk:
Needed to do 2 things
- add the new animTable to the array (this is what was causing the crash!)
- serialise generator.logicName member

Also fixed the way the new settings are accessed, so they can be overridden via modconf


luciusDXL (approved):
Looks good, thanks for the fix.

PR #480, opened 2024-12-11
Add svg redraw of app icon following modern icon guidelines

fpiesche:
This is a faithful vector redraw of the TFE icon following the best practices for app icon design recommended by e.g. Flathub (which implements the FreeDesktop standard) and Apple:

- The icon is on a square canvas but does not fill it completely (Flathub provide a guideline template for this)
- There is now a border around the black icon with a gradient matching the glyph, for better contrast on light and dark backgrounds alike
- The icon is saved as an optimized SVG, reducing its filesize to less than the existing .png icon


luciusDXL (approved):

luciusDXL:
Thanks, looks good.

PR #488, opened 2025-02-11
Fix building on Linux

sotpapathe:
Tested on Ubuntu 20.04 with CMake 3.16.3, GCC 9.4.0 and clang 10.0.0.


mlauss2:
Please update your distro instead.  Do you have a special need for ancient distros+tools?

luciusDXL (approved):

PR #493, opened 2025-02-26
Fix external data path mapping on linux machines

jerethk:
This is intended to fix the issue where the ext data files could not  be loaded from outside of the build directory on linux machines


jerethk:
This has been tested on both windows and linux and all appears to be working fine now.

devusb:
Can confirm also worked for me (Linux, NixOS 25.05).

luciusDXL (approved):

PR #496, opened 2025-03-07
Mod-conf overrides for various constants

jerethk:
This was requested from the community

The new overrides are
- Wall and floor damage
- Gas damage
- Gravity
- Battery consumption rates
- Supercharge duration

### DIFF
--- a/TheForceEngine/TFE_DarkForces/player.cpp
+++ b/TheForceEngine/TFE_DarkForces/player.cpp
@@ -3304,6 +3304,26 @@ namespace TFE_DarkForces
 		SERIALIZE(ObjState_OneHitCheats, s_oneHitKillEnabled, JFALSE);
 		SERIALIZE(ObjState_OneHitCheats, s_instaDeathEnabled, JFALSE);
 		SERIALIZE(ObjState_CrouchToggle, s_playerCrouch, 0);
+		SERIALIZE(ObjState_ConstOverrides, s_lowFloorDamage, PLAYER_DMG_FLOOR_LOW);
+		SERIALIZE(ObjState_ConstOverrides, s_highFloorDamage, PLAYER_DMG_FLOOR_HIGH);
+		SERIALIZE(ObjState_ConstOverrides, s_gasDamage, PLAYER_DMG_FLOOR_LOW);
+		SERIALIZE(ObjState_ConstOverrides, s_wallDamage, PLAYER_DMG_WALL);
+		SERIALIZE(ObjState_ConstOverrides, s_headlampConsumption, HEADLAMP_ENERGY_CONSUMPTION);
+		SERIALIZE(ObjState_ConstOverrides, s_gogglesConsumption, GOGGLES_ENERGY_CONSUMPTION);
+		SERIALIZE(ObjState_ConstOverrides, s_gasmaskConsumption, GASMASK_ENERGY_CONSUMPTION);
+		SERIALIZE(ObjState_ConstOverrides, s_shieldSuperchargeDuration, SUPERCHARGE_DURATION);
+		SERIALIZE(ObjState_ConstOverrides, s_weaponSuperchargeDuration, SUPERCHARGE_DURATION);
+		
+		s32 projectileGravity = 0;
+		if (serialization_getMode() == SMODE_WRITE)
+		{
+			projectileGravity = getProjectileGravityAccel();
+		}
+		SERIALIZE(ObjState_ConstOverrides, projectileGravity, FIXED(120));
+		if (serialization_getMode() == SMODE_READ)
+		{
+			setProjectileGravityAccel(projectileGravity);
+		}
 
 		s32 invSavedSize = 0;
 		if (serialization_getMode() == SMODE_WRITE && s_playerInvSaved)
--- a/TheForceEngine/TFE_DarkForces/player.cpp
+++ b/TheForceEngine/TFE_DarkForces/player.cpp
@@ -83,7 +83,7 @@ namespace TFE_DarkForces
 		HEADLAMP_ENERGY_CONSUMPTION = 0x111,    // fraction of energy consumed per second = ~0.004
 		GASMASK_ENERGY_CONSUMPTION  = 0x444,    // fraction of energy consumed per second = ~0.0167
 		GOGGLES_ENERGY_CONSUMPTION  = 0x444,    // fraction of energy consumed per second = ~0.0167
-		SUPERCHARGE_DURATION		= 6554,		// 45 seconds
+		SUPERCHARGE_DURATION        = 6554,     // 45 seconds
 	};
 
 	static const s32 c_pitchLimits[] =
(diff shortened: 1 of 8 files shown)


TheForceEngine/TFE_DarkForces/player.cpp
```diff
@@ -83,6 +83,7 @@ namespace TFE_DarkForces
 		HEADLAMP_ENERGY_CONSUMPTION = 0x111,    // fraction of energy consumed per second = ~0.004
 		GASMASK_ENERGY_CONSUMPTION  = 0x444,    // fraction of energy consumed per second = ~0.0167
 		GOGGLES_ENERGY_CONSUMPTION  = 0x444,    // fraction of energy consumed per second = ~0.0167
+		SUPERCHARGE_DURATION		= 6554,		// 45 seconds
```
ifeldshteyn, line 86: nitpick - can you align the value to the rest of the column? What are you tab settings? =)

TheForceEngine/TFE_DarkForces/player.cpp
```diff
@@ -780,6 +848,18 @@ namespace TFE_DarkForces
 		s_minEyeDistFromFloor = PLAYER_MIN_EYE_DIST_FLOOR;
 		s_gravityAccel        = PLAYER_GRAVITY_ACCEL;
 
+		// TFE - reset constants that may have been previously overridden
+		s_lowFloorDamage			= PLAYER_DMG_FLOOR_LOW;
```
ifeldshteyn, line 852: uniform spacing here if possible

TheForceEngine/TFE_DarkForces/player.cpp
```diff
@@ -183,6 +184,15 @@ namespace TFE_DarkForces
 	static angle14_32 s_playerObjYaw;
 	static RSector* s_playerObjSector;
 	static RWall* s_playerSlideWall;
+
+	// TFE - constants which can be overridden
+	static s32 s_lowFloorDamage = PLAYER_DMG_FLOOR_LOW;
+	static s32 s_highFloorDamage = PLAYER_DMG_FLOOR_HIGH;
+	static s32 s_gasDamage = PLAYER_DMG_FLOOR_LOW;
```
ifeldshteyn, line 191: So is gas damage just an alias for low floor damage?

TheForceEngine/TFE_DarkForces/projectile.cpp
```diff
@@ -83,6 +83,8 @@ namespace TFE_DarkForces
 	static fixed16_16 s_projNextPosZ;
 	static JBool      s_hitWater;
 	static RWall*     s_hitWall;
+
+	static s32		  s_projGravityAccel = PROJ_GRAVITY_ACCEL;		// TFE
```
ifeldshteyn, line 87: spacing

TheForceEngine/TFE_Settings/settings.cpp
```diff
@@ -1564,16 +1602,16 @@ namespace TFE_Settings
 
 		FilePath filePath;
 		FileStream file;
-		if (!TFE_Paths::getFilePath("MOD_CONF.txt", &filePath)) { return; }
+		if (!TFE_Paths::getFilePath("MOD_CONF.txt", &filePath) && !TFE_Paths::getFilePath("MOD_CONF.json", &filePath)) { return; }
```
ifeldshteyn, line 1605: It's a good idea to also support json but Remaster won't support it then. These will be TFE only changes.

ifeldshteyn (commented):
Looks good - just minor spacing issues

TheForceEngine/TFE_Settings/settings.cpp
jerethk: @luciusDXL what's your opinion about this?

TheForceEngine/TFE_DarkForces/player.cpp
jerethk: yeah - it's weird. In my VS they look aligned
image

I think it's because I have used tabs but @luciusDXL tends to just use spaces. I'll replace my tabs with spaces and see if that fixes it

TheForceEngine/TFE_DarkForces/player.cpp
```diff
@@ -573,6 +588,59 @@ namespace TFE_DarkForces
...
+		}
+		if (intMap.find("gravity") != intMap.end())
+		{
+			s_gravityAccel = FIXED(intMap["gravity"]);
+		}
+		if (intMap.find("projectileGravity") != intMap.end())
+		{
+			setProjectileGravityAccel(FIXED(intMap["projectileGravity"]));
+		}
+
```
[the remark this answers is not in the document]
jerethk: which lines look misaligned to you? Again, it all looks aligned in my IDE

TheForceEngine/TFE_DarkForces/player.cpp
jerethk: have replaced tabs with spaces here too, that should fix it

TheForceEngine/TFE_DarkForces/player.cpp
jerethk: I suspect the RE'd code just had plain numbers for all these things. And then @luciusDXL created a table of constants to make the code nicer to read & maintain. (Most likely, the original sourcecode also used constants, but these were lost in compilation)

Low floor damage and gas damage happen to be the same number.

TheForceEngine/TFE_DarkForces/projectile.cpp
jerethk: fixed

jerethk:
Thanks @ifeldshteyn . Github seems to render tabs differently from VS. It's annoying. But I fixed it just for you :-)

TheForceEngine/TFE_DarkForces/player.cpp
luciusDXL: To be clear, these values were stored in different memory addresses - so they were different variables in the original code. They just happen to use the same value.

TheForceEngine/TFE_Settings/settings.cpp
luciusDXL: TFE should support Remaster compatible mods. It is ok for mods to be TFE-only if needed or it doesn't make sense to run on the Remaster/vanilla though (such as structural differences or gameplay is too different).

jerethk: Yeah, so TFE is a superset of Remaster, and Remaster is (theoretically) a superset of vanilla.
Everything that runs on Remaster will run in TFE, but not vice versa.
By that reasoning, I think allowing either a TXT or JSON extension is reasonable.

Allowing it to be `.JSON` can help because peoples' text editors will automatically recognise JSON format from the extension, whereas this might not happen if the extension is `.TXT`

But happy to go either way :-)

luciusDXL: The issue is that it is another step - you have to remember to use txt instead of json otherwise you are wondering why your level doesn't work with the Remaster. So we should be consistent here. :)

In other words, we should make sure we are getting enough bang for the buck when introducing new points of failure. I don't think we get that here.

jerethk: Right, that makes sense. I've reverted it -- back to only taking the TXT extension

PR #497, opened 2025-03-10
Create player script class

jerethk:
@luciusDXL just a stub to begin with -- to check that you are happy with the approach before we do anything further.

All it has for now is the ability to get and set health.


jerethk:
Closing for now. Will do a PR when task is fully done

PR #504, opened 2025-03-23
Add GS_Player class

jerethk:
Introduces ability to manipulate the player via scripting

If this can be squeezed into the upcoming release, that would be terrific, but you don't think that's a good idea, that's fine, it can wait for the next one. It's your call to make :-) 
I understand that there are still the issues to sort out re save/load not working properly with scripting

### DIFF
--- a/TheForceEngine/TFE_DarkForces/Scripting/gs_player.cpp
+++ b/TheForceEngine/TFE_DarkForces/Scripting/gs_player.cpp
@@ -1,0 +1,424 @@
+#include "gs_player.h"
+#include "assert.h"
+#include <TFE_DarkForces/player.h>
+#include <TFE_DarkForces/mission.h>
+#include <TFE_DarkForces/pickup.h>
+#include <TFE_DarkForces/weapon.h>
+#include <TFE_System/system.h>
+#include <angelscript.h>
+
+namespace TFE_DarkForces
+{
+	void killPlayer()
+	{
+		sound_play(s_playerDeathSoundSource);
+		if (s_gasSectorTask)
+		{
+			task_free(s_gasSectorTask);
+		}
+		s_playerInfo.health = 0;
+		s_gasSectorTask = nullptr;
+		s_playerDying = JTRUE;
+		s_reviveTick = s_curTick + 436;
+	}
+
+	s32 getBatteryPercent()
+	{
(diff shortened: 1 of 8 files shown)


ifeldshteyn (commented):
looks good

TheForceEngine/TFE_DarkForces/Scripting/gs_player.cpp
```diff
@@ -0,0 +1,424 @@
...
+	}
+
+	void giveAutogun()
+	{
+		s_playerInfo.itemAutogun = JTRUE;
+	}
+
+	bool hasFusion()
+	{
+		return s_playerInfo.itemFusion == JTRUE;
```
luciusDXL, line 241: This should be `=` not `==`

TheForceEngine/TFE_DarkForces/Scripting/gs_player.cpp
```diff
@@ -0,0 +1,424 @@
...
+	{
+		if (s_wearingGasmask)
+		{
+			disableMask();
+		}
+		
+		s_playerInfo.itemMask = JFALSE;
+	}
+
+	bool hasPistol()
```
luciusDXL, line 209: I wonder if instead of having many get/set functions, it might be easier to have something like `giveWeapon(Weapon)`, etc. and then setup enums or something - so this API can be reduced to just a few functions (get/set per category probably).

luciusDXL (changes requested):
The == instead of = will need to be fixed before merging.

luciusDXL (approved):

TheForceEngine/TFE_DarkForces/Scripting/gs_player.cpp
jerethk: That is a sensible idea @luciusDXL 
Potentially we could have 

giveItem(RED_KEY)
giveItem(BLUE_KEY)
giveItem(MASK)
giveItem(DATATAPE)
...
setAmmo(DETONATORS, #)
setAmmo(ENERGY, #)

...
giveWEAPON(BLASTER)

and so on.

I won't document it on the wiki until we've made a final decision and updated it

luciusDXL: Yeah, I think that will be a lot cleaner - far fewer functions to worry about.

PR #512, opened 2025-03-26
Update gs_player

jerethk:
As discussed - use enums to reduce number of get/set functions


luciusDXL (approved):
This is much nicer, thank you.

PR #513, opened 2025-03-26
Harkov replay fix

ifeldshteyn:
1. Fixes harkov replay issue. 
2. Updated demo logs
3. Disable Fullbright cheat.


luciusDXL:
Thanks for the quick fix.

PR #517, opened 2025-04-02
Add ability to disable player movement/rotation/fire via scripting

jerethk:
The motive of this is to support the feature I am creating which allows the camera (eye) to be switched between the player object and other objects.
I want to be able to stop the player from moving while (for example) a cinematic sequence is playing and the eye is elsewhere.

I've allowed separate control over movement, rotation and firing so that modders can have fine-tuned control. Someone might want to suppress firing while still allowing movement, and so on.

### DIFF
--- a/TheForceEngine/TFE_DarkForces/Scripting/gs_player.cpp
+++ b/TheForceEngine/TFE_DarkForces/Scripting/gs_player.cpp
@@ -22,6 +22,21 @@ namespace TFE_DarkForces
 		AMMO_MISSILE,
 	};
 
+	enum ActionFlags : u32
+	{
+		ACTION_NONE        = 0u,
+		ACTION_MOVE        = FLAG_BIT(0),
+		ACTION_ROTATE      = FLAG_BIT(1),
+		ACTION_FIRE        = FLAG_BIT(2),
+		ACTION_ALL         = ACTION_MOVE | ACTION_ROTATE | ACTION_FIRE,
+		
+		// Others that could be implemented if desired?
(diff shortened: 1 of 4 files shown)


TheForceEngine/TFE_DarkForces/Scripting/gs_player.cpp
```diff
@@ -546,6 +546,65 @@ namespace TFE_DarkForces
 		}
 	}
 
+	void disableMovement()
```
luciusDXL, line 549: You can probably replace this with flags:
u32 s_playerActionFlags = 0; // or whatever
enum {
ACTION_MOVE = 1 << 0,
ACTION_ROTATE = 1 << 1,
ACTION_FIRE = 1 << 2,
};

And then have a few functions:
bool isPlayerActionEnabled(u32 flags) {}
void enablePlayerActions(u32 actions) {}
void disablePlayerActions(u32 actions) {}

The main reason is that, as it is now, you have to add multiple functions for every type of player action you want  to change. My thought, instead, is to just add a flag and keep the API to the same set of functions.

jerethk: Excellent suggestion, thank you very much!
(Should have thought of it myself)

I have done as you recommended with the scripting API.
In the game code, I'm leaving it as separate variables because I think that is easier and more readable there

luciusDXL (approved):

luciusDXL:
This greatly simplified the script API, thanks.

PR #518, opened 2025-04-04
Allow user to load difficulty

ifeldshteyn:
Allow FS to load the current difficulty


TheForceEngine/TFE_DarkForces/Scripting/scriptElev.cpp
```diff
@@ -23,6 +23,14 @@ namespace TFE_DarkForces
...
 	}
 
+	s32 getElevSpeed(ScriptElev* elev)
+	{
+		if (!ScriptElev::isScriptElevValid(elev)) { return 0; }
+
+		InfElevator* data = (InfElevator*)allocator_getByIndex(s_infSerState.infElevators, elev->m_id);
+		return data->type == IELEV_ROTATE_WALL ? floor16(angleToFloat(data->speed)) : floor16(data->speed);
+	}
+
```
jerethk, line 33: Thanks for adding this Karjala.
I have realised a (reasonably minor) flaw in my Elevator speed setter. Namely, in INF, it is possible to set an elevator to a floating point speed (like 6.5) whereas I have forced it to be integer (s32). 

@luciusDXL and @ifeldshteyn Do you think we should correct this, and allow getting/setting of speed as a float via scripting? If so, would you be able to please make that amendment with this PR?

TheForceEngine/TFE_DarkForces/Scripting/scriptElev.cpp
```diff
@@ -37,13 +45,15 @@ namespace TFE_DarkForces
...
 		// Variables
 		ScriptMemberVariable("int id", m_id);
 		// Functions
 		ScriptObjFunc("bool isValid()", isScriptElevValid);
 		// Properties
 		ScriptPropertyGetFunc("bool get_master()", getElevMaster);
-		
+
+		ScriptPropertyGetFunc("int get_speed()", getElevSpeed);
```
jerethk, line 56: If you decide to make the change, then this would become
`float get_speed()`
and 
`void set_speed(float)`

TheForceEngine/TheForceEngine.vcxproj
```diff
@@ -45,54 +45,54 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
```
jerethk, line 48: Oooh, I wonder if Lucius will approve this. I have it as v143 too, it would save me the hassle of continually stashing and unstashing this file whenever I merge/rebase etc.
I assumed he must have some reason to want to keep it at v141....

TheForceEngine/crashdump.dmp
jerethk, line 1: I don't think this is meant to be here? Mistake?

TheForceEngine/x64/Debug/TheForceEngine.exe.recipe
```diff
@@ -0,0 +1,11 @@
...
+<Project>
+  <ProjectOutputs>
+    <ProjectOutput>
+      <FullPath>C:\Users\Karjala\Source\Repos\TheForceEngine\x64\Debug\TheForceEngine.exe</FullPath>
+    </ProjectOutput>
+  </ProjectOutputs>
+  <ContentFiles />
+  <SatelliteDlls />
+  <NonRecipeFileRefs />
+</Project>
```
jerethk, line 11: I don't think these recipe files are supposed to be committed?

TheForceEngine/TheForceEngine.vcxproj
ifeldshteyn: CRUD - accidental commit =)

jerethk: lol

luciusDXL:
Currently VS2017 is used to preserve support for Win7 and 8 (theoretically at least).

Gerwin2k:
> Currently VS2017 is used to preserve support for Win7 and 8 (theoretically at least).

The ForceEngine Github builds work in Windows 7 x64 so far, practically that is.

TheForceEngine/TFE_DarkForces/Scripting/scriptElev.cpp
luciusDXL: Yeah, non-integer speeds are used in mods and even vanilla levels. So it would make sense to make it float instead of int.

jerethk: That was me with my brain not being fully engaged when I made the original addition :-)
If you can be bothered Karjala, you can make the amendment, otherwise I'll raise a change after this is merged.

TheForceEngine/TFE_DarkForces/Scripting/scriptElev.cpp
luciusDXL: This function should return a float - both angles and speed are fixed point.

jerethk: @ifeldshteyn @luciusDXL 
Ok I'm pretty much responsible for creating this issue, so here is the fix

https://github.com/luciusDXL/TheForceEngine/commit/2320077b47607f8cc4b988473fe1e186a7731ddf

Karjala, please go ahead and cherry pick it.

TheForceEngine/TFE_DarkForces/Scripting/scriptElev.cpp
```diff
@@ -23,12 +23,20 @@ namespace TFE_DarkForces
...
+		return data->type == IELEV_ROTATE_WALL ? angleToFloat(data->speed >> FRAC_BITS_16) : fixed16ToFloat(data->speed);
+	}
+
+	void setElevSpeed(float value, ScriptElev* elev)
 	{
 		if (!ScriptElev::isScriptElevValid(elev)) { return; }
 
 		InfElevator* data = (InfElevator*)allocator_getByIndex(s_infSerState.infElevators, elev->m_id);
-		data->speed = data->type == IELEV_ROTATE_WALL ? FIXED(floatToAngle(value)) : FIXED(value);
+		data->speed = data->type == IELEV_ROTATE_WALL ? intToFixed16(floatToAngle(value)) : floatToFixed16(value);
```
luciusDXL, line 39: Same issue here, you are losing the fractional part of the angular speed. angle degrees to DF angles -> float to fixed I think is the order here.

TheForceEngine/TFE_DarkForces/Scripting/scriptElev.cpp
jerethk: I'm a bit confused, I think you will need to explain in more detail please.

Let's say we want to set a speed of 20.5 degrees / sec
the formula currently there is `intToFixed16(floatToAngle(20.5))`

floatToAngle(20.5) = 20.5 * 45.51 = 932
intToFixed(932) = 932 << 16 = 61079552 

That seems correct, doesn't it? It looks the same as what is done when INF is parsed.

Then back the other way, formula is
`angleToFloat(61079552 >> FRAC_BITS_16)`

61079552 >> FRAC_BITS_16 = 932
angleToFloat(932) = 932 / 45.51 = 20.479

A small amount of precision has been lost but that's inevitable, isn't it?

TheForceEngine/TFE_DarkForces/Scripting/scriptElev.cpp
jerethk: Ok, I think see what you are saying. But line 2157 of `infSystem.cpp` uses `intToFixed` and therefore loses the precision as well. That is from vanilla, isn't it? Isn't it reasonable to just copy what vanilla does when it parses INF?

TheForceEngine/TFE_DarkForces/Scripting/scriptElev.cpp
```diff
@@ -23,12 +23,20 @@ namespace TFE_DarkForces
 		return (data->updateFlags & ELEV_MASTER_ON) != 0;
 	}
 
-	void setElevSpeed(s32 value, ScriptElev* elev)
+	float getElevSpeed(ScriptElev* elev)
+	{
+		if (!ScriptElev::isScriptElevValid(elev)) { return 0; }
+
+		InfElevator* data = (InfElevator*)allocator_getByIndex(s_infSerState.infElevators, elev->m_id);
+		return data->type == IELEV_ROTATE_WALL ? degreesToFixedAngle(data->speed) : fixed16ToFloat(data->speed);
```
jerethk, line 31: Just looking at this again -- I think you've got these round the wrong way.
This is the getter, so it should call `fixedAngleToDegrees`

TheForceEngine/TFE_DarkForces/Scripting/scriptElev.cpp
```diff
@@ -23,12 +23,20 @@ namespace TFE_DarkForces
...
+	}
+
+	void setElevSpeed(float value, ScriptElev* elev)
 	{
 		if (!ScriptElev::isScriptElevValid(elev)) { return; }
 
 		InfElevator* data = (InfElevator*)allocator_getByIndex(s_infSerState.infElevators, elev->m_id);
-		data->speed = data->type == IELEV_ROTATE_WALL ? FIXED(floatToAngle(value)) : FIXED(value);
+		data->speed = data->type == IELEV_ROTATE_WALL ? fixedAngleToDegrees(value) : floatToFixed16(value);
```
jerethk, line 39: This is the setter, so I think it should call `degreesToFixedAngle`

jerethk (changes requested):
Just looked at this again. A couple issues, I think

TheForceEngine/TFE_DarkForces/Scripting/scriptElev.cpp
jerethk: @Karjala22  I've built this and tested it and yes, you need to swap these functions around

Here it should be `fixedAngleToDegrees` 
And in the setter below it should be `degreesToFixedAngle`

jerethk (changes requested):

PR #520, opened 2025-04-06
Introduce ScriptObject class for object scripting

jerethk:
For this to work, I have created a "master list" of all objects that ever exist in a level: `s_objectRefList`
Objects are added to this list as they are created
When objects are deleted/freed, they are marked as removed but their position in the list is retained, so that IDs remain constant
(For now, I am not adding effects/explosions to the list, to keep down the size of the list. I have put the necessary code in place but commented it out. This is something that can be re-visited later if needed.)

Objects can be given a name and these names are serialised (along with the whole list)
The scripting API just has some basic functionality in it, that can be added to later


TheForceEngine/TFE_DarkForces/Scripting/gs_level.cpp
```diff
@@ -159,16 +161,77 @@ namespace TFE_DarkForces
...
+				break;
+			}
+		}
+		
+		ScriptObject object(objectId);
+		return object;
+	}
+
+	// Objects will be contained in the results array. Return value is the number of objects in the result.
+	int GS_Level::getAllObjectsByName(std::string name, CScriptArray& results)
```
luciusDXL, line 200: The array can already be queried for length in the script (arr.length) - so I don't think we need to also return the count.

TheForceEngine/TFE_DarkForces/Scripting/gs_level.cpp
```diff
@@ -159,16 +161,77 @@ namespace TFE_DarkForces
...
+		}
+		
+		ScriptObject object(id);
+		return object;
+	}
+
+	ScriptObject GS_Level::getObjectByName(std::string name)
+	{
+		const char* cname = name.c_str();
+		if (!cname || strcmp(cname, "") == 0 || s_objectRefList.empty())
```
luciusDXL, line 179: You can do `name.empty()` to see if the string is zero-length.

TheForceEngine/TFE_DarkForces/Scripting/scriptObject.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+	ScriptSector getSector(ScriptObject* sObject)
+	{
+		if (doesObjectExist(sObject))
+		{
+			SecObject* obj = s_objectRefList[sObject->m_id].object;
+			if (obj && obj->sector)
+			{
+				RSector* sector = s_levelState.sectors;
+				for (u32 i = 0; i < s_levelState.sectorCount; i++, sector++)
```
luciusDXL, line 62: You can just check sector->index and avoid the loop.
(You can still compare pointers if you want for extra security, but sector->index should never change) in the same level.

TheForceEngine/TFE_DarkForces/Scripting/gs_level.cpp
jerethk: Yep fair enough. Will change it.

TheForceEngine/TFE_DarkForces/Scripting/scriptObject.cpp
jerethk: Ah, so sector->index is always going to match the ScriptSector m_id?
If that's the case, that makes it much easier.

TheForceEngine/TFE_DarkForces/Scripting/scriptObject.cpp
```diff
@@ -0,0 +1,306 @@
...
+		ScriptPropertySetFunc("void set_radius(float)", setWorldWidth);
+		ScriptPropertySetFunc("void set_height(float)", setWorldHeight);
+		ScriptPropertySetFunc("void set_worldWidth(float)", setWorldWidth);
+		ScriptPropertySetFunc("void set_worldHeight(float)", setWorldHeight);
+		ScriptObjFunc("void matchPosition(Object)", matchPositionToTargetObj);
+		ScriptObjFunc("void matchPosition(string)", matchPositionToTargetObjByName);
+		
+		// Other functions
+		ScriptObjFunc("void delete()", deleteObject);
+		ScriptObjFunc("void addLogic(string)", addLogicToObject);
```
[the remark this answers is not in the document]
jerethk: @ifeldshteyn yes dealing with the logics will be in phase 2 
:-)

TheForceEngine/TFE_Jedi/Level/robjData.h
```diff
@@ -64,7 +64,21 @@ enum ObjStateVersion : u32
 	ObjState_CrouchToggle = 5,
 	ObjState_CustomLogics = 6,
 	ObjState_ConstOverrides = 7,
-	ObjState_CurVersion = ObjState_ConstOverrides,
+	ObjState_RefList = 8,
```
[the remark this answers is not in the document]
jerethk: If the save was created before the new object reference list existed, then you won't be able to load it, no.
The list will simply be empty.

TheForceEngine/TFE_DarkForces/Actor/actor.cpp
```diff
@@ -658,6 +658,8 @@ namespace TFE_DarkForces
 					corpse->worldWidth = 0;
 					corpse->entityFlags |= ETFLAG_CORPSE;
 					sector_addObject(obj->sector, corpse);
+
+					obj_addToRefList(corpse, ObjRefType_Corpse);	// scripting
```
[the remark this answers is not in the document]
jerethk: Because there are different types of objects (corpse, spawned item, projectile, etc.) and once you are inside `sector_addObject` you have no idea what type of object it is.
See the ObjectRefType enum

The idea is that eventually there may be functionality to let you get all objects of a certain category so you can do something with them.
For example make all projectiles stop in mid-air. Make all corpses levitate. Or whatever.

TheForceEngine/TFE_DarkForces/Scripting/scriptObject.cpp
```diff
@@ -0,0 +1,304 @@
...
+		{
+			obj_setSpriteAnim(obj);
+		}
+		else if (logicId >= KW_TROOP && logicId <= KW_SCENERY)
+		{
+			obj_setEnemyLogic(obj, logicId);
+		}
+		else if (logicId >= KW_BATTERY && logicId <= KW_AUTOGUN)
+		{
+			// TODO
```
luciusDXL, line 262: You might want to write out a message that the requested logic isn't supported or something like that.

jerethk: My intention is to get it done in the near future :-)

jerethk: OK I've done it now :-) 
Just a 1 line commit

Issue #521, opened 2025-04-08
Using ALT+F9 resets the whole mission if you do not have a quicksave


ifeldshteyn:
Issue is in the title. It should not doing anything if you don't have a quicksave.tfe

PR #522, opened 2025-04-08
Fix for quickload bug #521

ifeldshteyn:
This is a fix for #521 . You will no longer reset the mission if you are missing the quicksave.


luciusDXL (approved):

luciusDXL:
Thanks for the fix.

PR #526, opened 2025-04-13
Enable sending messages to a sector

jerethk:
Just looking for your opinion at this stage
It works, and GOTO_STOP appears to work fine, but I am nervous about possible unwanted side effects of assigning a value to the ` s_msgArg1` global variable. Can we be confident that this will  not break the messaging system? I have not tested comprehensively.

Would it be safer to only allow parameterless messages (like NEXT_STOP and PREV_STOP) until we can be sure about this?

### DIFF
--- a/TheForceEngine/TFE_DarkForces/Scripting/gs_level.cpp
+++ b/TheForceEngine/TFE_DarkForces/Scripting/gs_level.cpp
@@ -159,6 +159,18 @@ namespace TFE_DarkForces
 		setProjectileGravityAccel(FIXED(pGrav));
 	}
 
+	void sendMessageToSector(ScriptSector sSector, MessageType messageType, u32 evt, u32 msgArg1 = 0)
+	{
+		if (!isScriptSectorValid(&sSector))
+		{
+			return;
+		}
+
+		s_msgArg1 = msgArg1;
+		RSector* sector = &s_levelState.sectors[sSector.m_id];
+		message_sendToSector(sector, nullptr, evt, messageType);
+	}
+
 	bool GS_Level::scriptRegister(ScriptAPI api)
 	{
 		ScriptElev scriptElev;
@@ -225,6 +237,16 @@ namespace TFE_DarkForces
 			ScriptEnumStr(SECTORPROP_CEIL_TEX);
 			ScriptEnumStr(SECTORPROP_AMBIENT);
 
+			ScriptEnumRegister("MessageType");
+			ScriptEnum("M_TRIGGER",   MSG_TRIGGER);
+			ScriptEnum("NEXT_STOP",   MSG_NEXT_STOP);
+			ScriptEnum("PREV_STOP",   MSG_PREV_STOP);
+			ScriptEnum("GOTO_STOP",   MSG_GOTO_STOP);
+			ScriptEnum("DONE",        MSG_DONE);
+			ScriptEnum("WAKEUP",      MSG_WAKEUP);
+			ScriptEnum("MASTER_ON",   MSG_MASTER_ON);
+			ScriptEnum("MASTER_OFF",  MSG_MASTER_OFF);
+
 			// Functions
 			ScriptObjMethod("Sector getSector(int)", getSectorById);
 			ScriptObjMethod("Elevator getElevator(int)", getElevator);
@@ -232,6 +254,9 @@ namespace TFE_DarkForces
 
 			ScriptPropertySet("void set_gravity(int)", setGravity);
 			ScriptPropertySet("void set_projectileGravity(int)", setProjectileGravity);
+
+			ScriptObjFunc("void sendMessage(Sector, int, uint, uint)", sendMessageToSector);
+			ScriptObjFunc("void sendMessage(Sector, int, uint)", sendMessageToSector);
 
 			// -- Getters --
 			ScriptLambdaPropertyGet("int get_minLayer()", s32, { return s_levelState.minLayer; });


TheForceEngine/TFE_DarkForces/Scripting/gs_level.cpp
```diff
@@ -159,6 +159,18 @@ namespace TFE_DarkForces
...
 	}
 
+	void sendMessageToSector(ScriptSector sSector, MessageType messageType, u32 evt, u32 msgArg1 = 0)
+	{
+		if (!isScriptSectorValid(&sSector))
+		{
+			return;
+		}
+
+		s_msgArg1 = msgArg1;
```
ifeldshteyn, line 169: This is used throughout the codebase - mostly on weapon data for both player and phase one. I guess what matters is where the FS gets executed compared to other checks of this variable. I'd fully test this (preferably with phase1s in your test map).

jerethk: Yes that's why I'm nervous

jerethk:
The other question is, should this be a function on ScriptSector instead.
So,
Sector.sendMessage(msgType, param)

TheForceEngine/TFE_DarkForces/Scripting/gs_level.cpp
luciusDXL: We shouldn't rely on script versus task execution order I think. So the best bet here might be to create a message queue and have the messages sent during the elevator task function (which is where stop messages are handled). 

Even though script calls are made from this function, they are actually queued up as script threads with zero delay.

luciusDXL:
> The other question is, should this be a function on ScriptSector instead. So, Sector.sendMessage(msgType, param)
> 
> Might be easier than level.sendMessage(Sector, msgType, param)

I agree with this - it would make more sense from a scripter's point of view since the message is being sent to the sector.

TheForceEngine/TFE_DarkForces/Scripting/gs_level.cpp
jerethk: Probably best to leave this out initially, until it can be engineered in a "thread safe" way.
In the first instance, we can just allow for parameterless messages like "NEXT_STOP"

The `s_msArg1` is used very widely, to send damage messages, control the weapon system, etc. so best not to mess with it until we can be sure

jerethk:
> > The other question is, should this be a function on ScriptSector instead. So, Sector.sendMessage(msgType, param)
> > Might be easier than level.sendMessage(Sector, msgType, param)
> 
> I agree with this - it would make more sense from a scripter's point of view since the message is being sent to the sector.

Unfortunately that creates a drawback because of way the ScriptSector is passed in as the final argument. It means that overloading the function will not work.

If we keep it as a gs_level function we can have this
```
void sendMessage(Sector sector, int messageType, uint event, uint param)"
void sendMessage(Sector sector, in tmessageType, uint event)"
void sendMessage(Sector sector, int messageType)"
```

So you can do all of the following
sendMessage(NEXT_STOP)
sendMessage(NEXT_STOP, 131072)
sendMessage(GOTO_STOP, 131702, 3)

This doesn't work (afaics) if  the function is called on a ScriptSector
If you have these function signatures
```
void sendMessage(int messageType)
void sendMessage(int messageType, uint event)
```
And you do this
`mySector.sendMessage(NEXT_STOP)`

Angelscript chooses the second signature and assigns the ScriptSector to `uint event` (resulting in a crash)

So we are forced to require all 3 parameters (messageType, event, param) and the scripter will just have to set the 2nd and 3rd params to 0 if they don't want to use them.

Opinions?

ifeldshteyn:
Default values are fine just as long as there is no ambiguity between the default value and an expected value (Ex: you really want to pass along a 0). What will you use as default?

jerethk:
> Default values are fine just as long as there is no ambiguity between the default value and an expected value (Ex: you really want to pass along a 0). What will you use as default?

the issue is not default values. It is that an overloaded function does not work. 

It means that a scripter will have to write
`mySector.sendMessage(NEXT_STOP, 0, 0)` to send a next_stop message to an elevator with no further arguments

Whereas if it is a function on gs_level, overloading is possible so they can do
`level.sendMessage(mySector, NEXT_STOP)`

I don't have a super strong preference either way, but the latter is more similar to INF (with which people are already familiar) where arguments are optional

eg.
message: 1 mySector NEXT_STOP
message: 1 mySector NEXT_STOP 131072

Does that make sense

luciusDXL:
> > > The other question is, should this be a function on ScriptSector instead. So, Sector.sendMessage(msgType, param)
> > > Might be easier than level.sendMessage(Sector, msgType, param)
> > 
> > 
> > I agree with this - it would make more sense from a scripter's point of view since the message is being sent to the sector.
> 
> Unfortunately that creates a drawback because of way the ScriptSector is passed in as the final argument. It means that overloading the function will not work.
> 
> If we keep it as a gs_level function we can have this
> 
> ```
> void sendMessage(Sector sector, int messageType, uint event, uint param)"
> void sendMessage(Sector sector, in tmessageType, uint event)"
> void sendMessage(Sector sector, int messageType)"
> ```
> 
> So you can do all of the following in forcescript sendMessage(NEXT_STOP) sendMessage(NEXT_STOP, 131072) sendMessage(GOTO_STOP, 131702, 3)
> 
> This doesn't work (afaics) if the function is called on a ScriptSector If you have these function signatures
> 
> ```
> void sendMessage(int messageType)
> void sendMessage(int messageType, uint event)
> ```
> 
> And you do this `mySector.sendMessage(NEXT_STOP)`
> 
> Angelscript chooses the second signature and assigns the ScriptSector to `uint event` (resulting in a crash)
> 
> So we are forced to require all 3 parameters (messageType, event, param) and the scripter will just have to set the 2nd and 3rd params to 0 if they don't want to use them.
> 
> Opinions?

Sure, in that case keeping the functionality in level is fine. The point is to avoid unnecessary complexity, especially for the scripter, and being able to support function overloads naturally trumps the slight improvement for having this in Sector.

TheForceEngine/TFE_DarkForces/Scripting/gs_level.cpp
luciusDXL: I think once a "message queue" is added for this, than all of the sector messages would use it - rather than using two different systems with potentially slightly different timing. But if you want to do it one way for now and change it later that is fine.

jerethk: I suppose it depends on how far off the "message queue" would be, vs. how far off till you want to announce the forcescript system as "mature" enough for use.

PR #539, opened 2025-06-06
Attack mod overrides

ifeldshteyn:
This is a POC - projectile offset  spawning code.  Now you can tell the game exactly where to spawn the projectile. This will work no matter where the actor is facing. 

// Projectile spawn details guide.
// 
// Z value positive =  projectile spawns to the LEFT of actor
// Z value negative =  projectile spawns to the RIGHT of actor 
// X value postivie =  projectile spawns in FRONT of actor
// X value negative =  projectile spawns BEHIND the actor
// Y value positive =  projectile spawns BELOW the actor
// Y value negative =  projectile spawns ABOVE the actor
//

I am adding an example logic offset. The stormtrooper cannot move but the rockets will spawn depending the offset values. 

      "fireOffset": [
          0,
          -4,
          5
      ],

[OFFSET.zip](https://github.com/user-attachments/files/20634877/OFFSET.zip)


ifeldshteyn:
Note @luciusDXL @jerethk I think we need to up the serialization version for the additional variable?

ifeldshteyn:
It was merged with jereth's PR

PR #542, opened 2025-06-07
Colors POC

ifeldshteyn:
POC  - allow you to set colors for all automaps.


TheForceEngine/TFE_DarkForces/automap.cpp
```diff
@@ -489,32 +475,90 @@ namespace TFE_DarkForces
...
 	u8 automap_getWallColor(RWall* wall)
 	{
 		u8 color;
+		// Default thickness - GPU only 
+		setLineThickness(1.5f);
+
 		if (wall->flags1 & WF1_HIDE_ON_MAP)
 		{
-			color = WCOLOR_INVISIBLE;
+			color = TFE_Settings::getMapColorByName("Invisible Wall")->colorMapIndex;
```
luciusDXL, line 507: I don't really like the idea of doing value lookups by string for every wall when drawing a map. I think there should be a table that is directly indexed that is filled out based on the settings instead.

ifeldshteyn:
I am closing this. We can revisit later.

PR #570, opened 2025-09-18
Secret count

ifeldshteyn:
Add a way to show total # of secrets and allow for centering of the HUD data.

image

image


TheForceEngine/TFE_Settings/settings.cpp
```diff
@@ -554,29 +554,31 @@ namespace TFE_Settings
...
+		writeKeyValue_Bool(settings, "showReplayCounter", s_gameSettings.df_showReplayCounter);
+		writeKeyValue_Int(settings,  "recordFrameRate", s_gameSettings.df_recordFrameRate);
+		writeKeyValue_Int(settings,  "playbackFrameRate", s_gameSettings.df_playbackFrameRate);
+		writeKeyValue_Bool(settings, "enableRecording", s_gameSettings.df_enableRecording);
+		writeKeyValue_Bool(settings, "enableRecordingAll", s_gameSettings.df_enableRecordingAll);
+		writeKeyValue_Bool(settings, "demologging", s_gameSettings.df_demologging);
+		writeKeyValue_Bool(settings, "autoNextMission", s_gameSettings.df_autoEndMission);
+		writeKeyValue_Bool(settings, "showKeyUsed", s_gameSettings.df_showKeyUsed);
+		writeKeyValue_Bool(settings, "showKeyColors", s_gameSettings.df_showKeyColors);
+		writeKeyValue_Bool(settings, "centerHudPos", s_gameSettings.df_centerHudPosition);
```
jerethk, line 581: @ifeldshteyn @luciusDXL I wonder if, rather than an endless proliferation of new settings (which could overwhelm players), it would be smarter to group them together.

For example, any future logic-related enhancements I add, I will roll into the custom logics setting so that they all enable or disable together. I don't consider it likely that the majority of people will want to have ultra granular control over such things.

Perhaps "show secret count" could be rolled into "show secret found message", and renamed as "show secrets found" or something like that.

Likewise, "show key used" and "show key colours" I think could be rolled into a single setting.

What do you guys think.

TheForceEngine/TheForceEngine.vcxproj
```diff
@@ -39,60 +39,60 @@
     <ProjectGuid>{31CD2991-9EF0-4AA9-AEEB-BA1194D2DAB4}</ProjectGuid>
```
luciusDXL, line 39: This file should be excluded.

ifeldshteyn:
Added merge fix - seems to build now.

ifeldshteyn (commented):
Can this be rechecked ? I am blocked from committing other stuff @luciusDXL @jerethk

luciusDXL (approved):

PR #473, opened 2024-11-21
Add support for compiling on macOS

apancik:
This pull request introduces changes to enable compiling The Force Engine on macOS. It's a cleanup of #472 (thank you @mlauss2 for your help!)

Hardware: Intel Mac / macOS Sequoia 15.1 / XCode 16.1
SDL2: version 2.30.9, SDL2_image: version 2.8.2, RtMidi: version 6.0.0

Screenshot 2024-11-19 at 5 57 22 PM
Screenshot 2024-11-19 at 5 57 44 PM
Screenshot 2024-11-19 at 5 58 06 PM

PR #584, opened 2025-12-26
macOS Support

tomkidd:
This PR represents the changes I made to create the build I use on Mac Source Ports

https://www.macsourceports.com/game/swdf

As mentioned in the comments, it's a combination of PR #473 from @apancik, [this comment](https://github.com/luciusDXL/TheForceEngine/issues/90#issuecomment-2816098327) from @gilmorem560 about updating AngelScript, and a couple of additional tweaks. 

It might be better to rebase the original PR (I want to make sure credit is given where it's due), and one of the tweaks is App Bundle specific so I'm not sure if that's the best way to handle path names, but in any event here's my contribution upstream.

### DIFF
--- a/TheForceEngine/TFE_FileSystem/paths-posix.cpp
+++ b/TheForceEngine/TFE_FileSystem/paths-posix.cpp
@@ -183,7 +183,11 @@ namespace TFE_Paths
 	{
 		char p[TFE_MAX_PATH];
 		memset(p, 0, TFE_MAX_PATH);
+#ifdef __APPLE__
+		FileUtil::getExecutionDirectory(p);
+#else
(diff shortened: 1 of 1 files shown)


apancik:
Given that #473 has been open for about a year, I’m not sure where it sits in the maintainers’ queue.

Just in case, I’ve rebased and force-pushed the PR branch of #473. If you rebase my branch onto yours, we can preserve the history. Then I can close my PR and we can keep this one up. I can do the rebase too, but I’d prefer you do it so GitHub doesn’t end up attributing me as a co-author on your commits.

I tested that flow locally and it worked: everything compiles cleanly.

PR #591, opened 2026-02-11
Allow non 64x64 textures on floor and ceiling in floating point renderer

jerethk:
Required changes are very simple. Floating point renderer will now match GPU.
Original behaviour in fixed point renderer is preserved.


ifeldshteyn (approved):
Looks good - Ben should be happy.

PR #592, opened 2026-02-15
TD & Mine animation fix (proposed)

jerethk:
This is a subtle bug that has been present since the early days - it is seen when you use your final TD or mine.

@luciusDXL could you please compare with your original RE code?
I have reasoned out a fix which seems to give correct behaviour (cf. DOS), but I don't have access to the original code to check it. - thanks
It would be ideal if we can match the original code.

### DIFF
--- a/TheForceEngine/TFE_DarkForces/weapon.cpp
+++ b/TheForceEngine/TFE_DarkForces/weapon.cpp
@@ -485,6 +485,10 @@ namespace TFE_DarkForces
 		{
 			weapon->frame = 2;
 		}
+		else
+		{
+			weapon->frame = 0;
+		}
 	}
 
 	void weapon_computeMatrix(fixed16_16* mtx, angle14_32 pitch, angle14_32 yaw)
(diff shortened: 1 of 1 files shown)


TheForceEngine/TFE_DarkForces/weapon.cpp
```diff
@@ -1020,10 +1024,6 @@ namespace TFE_DarkForces
 			
 			const u8* atten = RClassic_Fixed::computeLighting(weaponLightingZDist, 0);
 			TextureData* tex = weapon->frames[weapon->frame];
-			if (weapon->ammo && *weapon->ammo == 0 && (weapon->ammo == &s_playerInfo.ammoDetonator || weapon->ammo == &s_playerInfo.ammoMine))
```
ifeldshteyn, line 1023: Isn't this code to reset to Fist if you are out of ammo? @jerethk

jerethk: I have a feeling Lucius may have added it as a temporary hack.
It doesn't give correct behaviour.
Look closely as you throw the last TD or mine in your inventory. The mine/TD vanishes incorrectly as soon as you start the action, rather than when the action is finished

ifeldshteyn: I remember this bug - so does this code removal fixes it?

jerethk: The change as a whole fixes it - question is, whether it is true to the RE

PR #595, opened 2026-02-28
Undo / redo for LevelNotes

acpushbutton:
Adds the functions and hooks to implement stepwise undo / redo for LevelNote objects.

It follows the structure and naming conventions of other object categories in the history system, extending enums as necessary, leveraging the existing read/writeLevelNoteFromSnapshot functions which already captures note data in full snapshots. It's a simple capture of the entire LevelNote vector, no optimisations for single item changes.

Includes logic to bundle sequential attribute updates into a single history step to reduce noise, like that from working in the Info Panel.

Some minor changes to editNotes.cpp to capture events correctly for the history system. It looks like this belongs in editTransform.cpp but decided against that to keep the scope reasonable.

### DIFF
--- a/TheForceEngine/TFE_Editor/LevelEditor/editNotes.cpp
+++ b/TheForceEngine/TFE_Editor/LevelEditor/editNotes.cpp
@@ -2,6 +2,7 @@
 #include "hotkeys.h"
 #include "levelEditor.h"
 #include "levelEditorData.h"
+#include "levelEditorHistory.h"
 #include "sharedState.h"
 #include <TFE_Editor/editorConfig.h>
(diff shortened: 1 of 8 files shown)


jerethk (approved):
Looks good to me, thank you for this addition.
I did notice that each snapshot holds the entire list rather than the a single note which had a delta, but this _seems_ ok. Each note is 44 bytes plus the string so not gigantic. Probably wouldn't blow up the RAM under most ordinary circumstances....

I have tested it locally and it works as expected.

ifeldshteyn:
This works - just tested it.

PR #596, opened 2026-04-01
Support logic scriptcall variables in editor

jerethk:
Now that logic scriptcalls have been merged, support should be added to the editor too.


ifeldshteyn (approved):
Ok for new script variables.

PR #599, opened 2026-05-03
Fix for camera pitch clamping (software renderer)

jerethk:
This fixes a bug associated with https://github.com/luciusDXL/TheForceEngine/pull/468
where I clamped the camera's pitch for the software renderer, but did not account for situations where pitch is expressed as a value between 180 and 360.

Angles between 180 and 360 need to first be converted to a value between -180 and 0, by subtracting 360
(This will not fix values that are > 540 and < 180 but I think it is fair to assume that such values won't be encountered in practice. This issue only occurred when transforming the camera object with a VUE, where all pitch values will be between 0 and 360)


ifeldshteyn (approved):
Looks good.

PR #600, opened 2026-05-03
Smooth VUEs amendment

jerethk:
The existing interpolation logic (introduced by https://github.com/luciusDXL/TheForceEngine/pull/288) does not take into account the wrapping-around of angles at 360 degrees. 
So, if an angle (eg. yaw) is 358 degrees in one frame and 2 degrees in the next frame, the interpolation will happen downwards from 358 to 2, instead of upwards (358, 359, 0, 1, 2).

This is not an issue with 3d-object VUEs, which ignore the angles and use the transformation matrix to render each keyframe. But it is noticed when using VUEs in camera mode, to move a camera or sprite around.

I think the solution to this is to test whether the delta is more than 180 degrees -- if so, interpolate via the shorter distance arc

### DIFF
--- a/TheForceEngine/TFE_DarkForces/vueLogic.cpp
+++ b/TheForceEngine/TFE_DarkForces/vueLogic.cpp
@@ -491,13 +491,24 @@ namespace TFE_DarkForces
 		{
 			interpolatedFrame->mtx[i] = lerp(previous->mtx[i], current->mtx[i], t);
 		}
+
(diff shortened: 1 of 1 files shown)


jerethk:
NB. I ran this past @kevinfoley on discord (he wrote the original VUE smoothing code) and he said he was ok with it.

ifeldshteyn (approved):
This looks good - I talked to Kevin and I understand the issue this is trying to solve.

PR #601, opened 2026-05-09
Added 2/4x supersampling

hurleybird:
My favorite way to play old 3D games is at a low resolution with supersampling. High resolutions tend to expose low asset quality, but low resolutions are marred by unseemly stair-stepping. Supersampling gives the best of both worlds.

Supersampled DF

Full disclosure, in case it matters: I used Codex to add the feature (I don't have much of any OpenGL experience to speak of). I've done some testing and due diligence (works with all the various graphics options that I could find, resizing the windows works, supersampling initially broke vanilla sky, but got it working). Nonetheless you may want to take that into account when looking things over.

### DIFF
--- a/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
+++ b/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
@@ -187,6 +187,20 @@ namespace TFE_FrontEndUI
 		"True Color",           // COLORMODE_TRUE_COLOR
 	};
 
+	static const char* c_supersampling[] =
+	{
+		"Off",
+		"2x",
+		"4x",
+	};
+
+	static const s32 c_supersamplingFactor[] =
+	{
+		1,
+		2,
+		4,
+	};
+
 	typedef void(*MenuItemSelected)();
 
 	static s32 s_resIndex = 0;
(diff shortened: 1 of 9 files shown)


mlauss2:
I tried this, but I see no visual difference at all with 2x or 4x (2560x1600 res, no bilinear filter).  How can I tell it's actually on?

hurleybird:
Try with a lower resolution, like 480 or 600p.

I think it looks best with 8-bit (I like interpolated, but either) non-dithered, since SSAA is able to clean up the stair stepping between texels.

mlauss2:
> Try with a lower resolution, like 480 or 600p.
> 
> I think it looks best with 8-bit (I like interpolated, but either) non-dithered, since SSAA is able to clean up the stair stepping between texels.

Thanks, I see it now.  Great work!  Over 800x600 it's barely noticeable, but makes a ton of difference if played at the original 320x200 resolution.

hurleybird:
> > Try with a lower resolution, like 480 or 600p.
> > I think it looks best with 8-bit (I like interpolated, but either) non-dithered, since SSAA is able to clean up the stair stepping between texels.
> 
> Thanks, I see it now. Great work! Over 800x600 it's barely noticeable, but makes a ton of difference if played at the original 320x200 resolution.

Thank you! It's visible to me even at 4K, but I have a big display and I'm sensitive to stair stepping and temporal noise. But yeah, outsized difference at low res (which is my preference for old 3D games)

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
```diff
@@ -3190,6 +3204,23 @@ namespace TFE_FrontEndUI
...
+			s32 supersamplingIndex = 0;
+			if (graphics->supersampling == 2)
+			{
+				supersamplingIndex = 1;
+			}
+			else if (graphics->supersampling == 4)
+			{
+				supersamplingIndex = 2;
+			}
+
```
mlauss2, line 3216: supersamplingIndex = graphics->supersampling >> 1   ?

mlauss2: What about e.g. 8x/16x ? would they make any meaningful visual difference?

mlauss2 (commented):
the rest looks okay to me, pretty clean and introduces no new build warnings.

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
hurleybird: Maybe, but I somewhat doubt it. I can't remember seeing anything do >4x before. I imagine that eventually you'd run into performance issues, and issues with exceeding max supported texture size on some GPUs.

ifeldshteyn:
Hi, I didn't see build checks for this PR. Was it disabled? @jerethk are we missing github action definions?

PR #603, opened 2026-06-01
Texture Flag Fix

ifeldshteyn:
Fix for texture flags not being preserved between saves as seen as Dark Tide 3 INVSPACE.BM bug. Adds compatibility for older  save files.

### DIFF
--- a/TheForceEngine/TFE_Jedi/Level/rtexture.cpp
+++ b/TheForceEngine/TFE_Jedi/Level/rtexture.cpp
@@ -12,6 +12,7 @@
 #include <TFE_Jedi/Serialization/serialization.h>
 #include <TFE_System/math.h>
 #include <TFE_Settings/settings.h>
+#include <TFE_Jedi/Level/levelData.h>
 #include <unordered_map>
 
 using namespace TFE_DarkForces;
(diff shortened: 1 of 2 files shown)


TheForceEngine/TFE_Jedi/Level/rtexture.cpp
```diff
@@ -189,14 +190,29 @@ namespace TFE_Jedi
...
 
 			// If reading, we need to load the texture now.
 			if (serialization_getMode() == SMODE_READ)
 			{
 				const char* name = list->name.c_str();
 				list->texture = bitmap_load(name, 1, POOL_LEVEL, false);
 				s_textureTable[POOL_LEVEL][name] = i;
+
+				// Restore the runtime flags that bitmap_load() masked away
+				if (list->texture && LevelState_TextureFlags >= serialization_getVersion())
```
jerethk: Have you got this around the wrong way?

Should it be `serialization_getVersion() >= LevelState_TextureFlags ` ?

i.e. you only want this done if the savefile version is **greater than or equal to** 4

jerethk (commented):
Might have found a problem, can you please check @ifeldshteyn

jerethk (approved):

PR #605, opened 2026-07-03
Use std::filesystem to get a standard temp directory

fpiesche:
This requires C++17, but will ensure the temporary directory is in a system-standard location by OS ([docs](https://en.cppreference.com/cpp/filesystem/temp_directory_path)). This also fixes mods that create temporary archives running in the Flatpak (where `PATH_PROGRAM_DATA` is read-only).


ifeldshteyn:
Missing filesystem import. Can we stick this into TFE_Paths ? Remember there are two versions NT and POSIX

PR #606, opened 2026-07-05
Level bounds / layers fixes

acpushbutton:
1. updateLevelBounds(): Updated to handle empty level case avoiding a crash.

2. Update undo, redo & history popup to refresh level bounds after replaying history.

3. updateCurrentLayer(): new function to make useful changes to active layer after undo / redo actions, effectively "Following" layer changes as it expands or shrinks.

### DIFF
--- a/TheForceEngine/TFE_Editor/LevelEditor/levelEditorData.cpp
+++ b/TheForceEngine/TFE_Editor/LevelEditor/levelEditorData.cpp
@@ -915,20 +915,28 @@ namespace LevelEditor
 	{
 		if (!sector)
 		{
-			s_level.bounds[0] = { FLT_MAX,  FLT_MAX,  FLT_MAX };
-			s_level.bounds[1] = { -FLT_MAX, -FLT_MAX, -FLT_MAX };
-			s_level.layerRange[0] = INT_MAX;
-			s_level.layerRange[1] = -INT_MAX;
 			const size_t count = s_level.sectors.size();
 			sector = s_level.sectors.data();
-			for (size_t i = 0; i < count; i++, sector++)
+
+			if (count == 0)
 			{
-				updateBoundsWithSector(sector);
+				s_level.bounds[0] = { 0 };
+				s_level.bounds[1] = { 0 };
+				s_level.layerRange[0] = 0;
+				s_level.layerRange[1] = 0;
 			}
-		}
-		else
-		{
-			updateBoundsWithSector(sector);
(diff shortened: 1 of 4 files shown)


TheForceEngine/TFE_Editor/LevelEditor/levelEditorData.cpp
```diff
@@ -915,20 +915,28 @@ namespace LevelEditor
...
+				s_level.layerRange[0] = INT_MAX;
+				s_level.layerRange[1] = -INT_MAX;
+
+				for (size_t i = 0; i < count; i++, sector++)
+				{
+					updateBoundsWithSector(sector);
+				}
 			}
-		}
-		else
```
jerethk, line 929: I think this `else` block should be re-instated.
All _current_ calls to this function appear to have `sector == nullptr` but it is obviously intended to be callable with a sector. (This might be implemented in future)

In that case, it will need the `else` block or else the function will just do nothing.

TheForceEngine/TFE_Editor/LevelEditor/levelEditorData.cpp
```diff
@@ -5489,4 +5497,28 @@ namespace LevelEditor
 			}
 		}
 	}
+
+	// Update visible layer when history replay changes the overall layer bounds.
+	void updateCurrentLayer(s32* oldLayers)
```
jerethk: It would be easier to understand what this parameter is if it is called `oldLayerRange`
i.e. it is an array of 2 numbers, a min and a max

`oldLayers` makes it sound like a full array of map layers, so potentially confusing.

TheForceEngine/TFE_Editor/LevelEditor/levelEditorHistory.cpp
```diff
@@ -119,14 +119,21 @@ namespace LevelEditor
 		
 	void levHistory_undo()
 	{
+		s32 oldLayers[2] = { s_level.layerRange[0], s_level.layerRange[1] };
```
jerethk: As above, this would be more easy to understand if it is called `oldLayerRange` instead of `oldLayers`

And same on line 132.

jerethk (changes requested):
This looks good, thank you.
Just a couple of minor change requests.

jerethk (approved):
Thanks for the addition!

ifeldshteyn (approved):
ok

PR #607, opened 2026-07-07
Enable burst-fire for NPC secondary fire

jerethk:
This will enable NPCs with melee attacks to have burst-fire on their projectile (ranged) attack.
It copies what was done for https://github.com/TheForceEngine/TheForceEngine/pull/589 onto the secondary fire pathway.

When an NPC/actor has a melee attack, the melee attack uses the primary attack pathway, and projectile attack will use the secondary pathway


ifeldshteyn (approved):
Ok makes sense - burst reset to new random #

PR #608, opened 2026-07-07
Forcescript cannot register both GS_Level and LS_Level because of shared name

jerethk:
Bendansie discovered this bug in Forcescript.

Both GS_Level and LS_Level register in angelscript as "Level"
Therefore, if you run the game followed by the editor (or vice versa) in the same session, this leads to a collision, and the second registration will fail.

This is one proposed solution, working with the assumption that we still only want a single instance of Forcescript to exist per application instance.
I have moved all the Enums out to a shared class.
LS_Level is then renamed to "ELevel" at registration (E for "editor") -- note: I am a bit dubious about this, consider it an interim solution

**Note**
The same problem exists with GS_System and LS_System (both registered as "System"). This will have to be solved separately.


TheForceEngine/TFE_ForceScript/ScriptAPI-Shared/scriptLevelShared.h
```diff
@@ -0,0 +1,38 @@
...
+enum WallPart
+{
+	WP_MIDDLE = 0,
+	WP_TOP,
+	WP_BOTTOM,
+	WP_SIGN,
+
+	// see levelEditorData.h
+	WP_MID = 0,
+	WP_BOT = 2,
```
ifeldshteyn, line 27: I'd rather we not have dupes for BOT/BOTTOM and MID/MIDDLE

PR #611, opened 2026-07-08
Caption Bug Fix

ifeldshteyn:
fix for caption crash


jerethk (approved):
Thank you for finding and fixing this.
Tested - works well.

PR #612, opened 2026-07-08
Harmonisation of InfEntityMask and EntityTypeFlags

jerethk:
`InfEntityMask` and `EntityTypeFlags` are equivalent to each other.
Therefore, it is less confusing if the enum names align in the code
eg. ENEMY --> AI_ACTOR
WEAPON --> PROJECTILE

I have added some comments to explain things.
Tiny behaviour change in custom logics -- the ETFLAG_HAS_GRAVITY flag should be set for all non-flying enemies. That way all ground-based custom logics will be guaranteed to align with the vanilla logics (stormtroopers, bossk, gamorrean, etc.) when modders are trying to do tricky INF stuff.


ifeldshteyn (approved):
ok this makes sense - lets use that extra bit

PR #615, opened 2026-07-09
Fix guidelines persisting between levels

acpushbutton:
s_level.guidelines wasn't being cleared when loading a GOB, only when loading a TFL. Adding this line next to a cluster of existing initialisation steps.


jerethk (approved):
Good fix.

PR #617, opened 2026-07-09
Fix build warnings

mlauss2:
Fix all the build warnings emitted by gcc:
- drop 2 "externs" on file-scope variables,
- replace %d/%l/.. with PRI* equivalents. gcc is espcially noisy wrt. printf format mismatches.
- use strftime() instead of the home-made equivalent. Again, printf() format specifier warnings.

With this the build is now warning-free with gcc-15+.

### DIFF
--- a/TheForceEngine/TFE_DarkForces/Scripting/gs_system.cpp
+++ b/TheForceEngine/TFE_DarkForces/Scripting/gs_system.cpp
@@ -8,6 +8,7 @@
 #include <TFE_ForceScript/float3.h>
 #include <TFE_ForceScript/Angelscript/add_on/scriptarray/scriptarray.h>
 #include <angelscript.h>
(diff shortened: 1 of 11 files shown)


jerethk:
Hi @mlauss2 
On my system (Windows, VS 2022) it does not resolve `PRIi32` without adding an explicit
`#include "inttypes.h"`
at the top of the file. Would you mind adding this please?

image

The other issue I'm seeing is that when `sprintf` is used with `PRIi64`, the IDE gets confused and issues a warning. It compiles ok.
So the upshot is that we seem to be trading warnings in one setting with warnings in another setting... 🙃
Any thoughts?

image

mlauss2:
Oh hang on, I see my mistake...

jerethk (approved):
All looking good now. Thanks!

PR #618, opened 2026-07-09
Avoid busy waiting on the main and the midi threads

### LINKED ISSUE #402: More CPU activity (even with OpenGL renderer) than dosbox

mlauss2:
This is the original #405 forward-ported to latest master.  All credit goes to @MIvanchev.

Its benefits still apply: in my playtesting CPU load with external MIDI is reduced from a constant >105% to <2%, i.e. the midi threads' cpu usage is no longer noticeable.  I have been running this since the original #405 was posted, without any ill-effects.

Supersedes #405
Fixes #402

### DIFF
--- a/TheForceEngine/TFE_Audio/midiPlayer.cpp
+++ b/TheForceEngine/TFE_Audio/midiPlayer.cpp
@@ -39,6 +39,26 @@ namespace TFE_MidiPlayer
 	{
 		MidiPlayerCmd cmd;
 		f32 newVolume;
+
+		static inline MidiCmd pause()
+		{
+			return { MIDI_PAUSE };
+		}
+
+		static inline MidiCmd resume()
+		{
+			return { MIDI_RESUME };
+		}
+
+		static inline MidiCmd stop()
+		{
+			return { MIDI_STOP_NOTES };
+		}
+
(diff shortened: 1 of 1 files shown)


jerethk:
Thanks @mlauss2  and @MIvanchev 
I'm not sure the reason @luciusDXL kept the previous PR aside. Possibly he just got busy and didn't get time to look at it. Will try to bring it to his attention.

ifeldshteyn:
I took a look yeah if this improves processing I am for it. Was Lucius too swamped to look at this in 2024?

mlauss2:
Without this, the midi thread is calling detectHangingNotes() as fast as it can, even if it's not necessary.  With the patch it sleeps until either the next beat or a command is issued from the main thread. This is independent of the chosen midi renderer (SF2/OPL3/External midi).

jerethk:
@mlauss2 hi, was there a reason this PR was deleted?

PR #619, opened 2026-07-09
Replace RtMidi with libremidi

### LINKED ISSUE #597: System MIDI Support on MacOS
Thanks for the MacOS support,

I can compile it and it runs great on M1 and M4 Macbook.
MIDI seems only to support SF2 or OPL3.
When choosing "System Midi" there is only the text "Disabled: no ports found".
I have real MIDI ports and virtual MIDI Ports in my system. 
Other Applications like DosBox can find and use them.
So I think at the moment this feature is not supported.
Is it planed to support it? 
I really want to connect my real Roland SC-88. 

Thanks in advanced

mlauss2:
RtMidi has been last updated in Mid-2023; libremidi is based on it and actively maintained.

Caveats:
- Currently this only works with CMake, as the CMakeFile does an automatic git pull, configuration and build of libremidi for the target platform (it's actually quite pleasant to integrate!)
It works quite well for me on Linux; however I have no idea how to integrate this with the visual studio project system.
- it needs a C++20 compiler. (gcc-11+, clang-8+, msvc-19+).

EDIT: reporter confirmed that it indeed fixes #597 

Comments welcome!


jerethk:
This looks good, thanks. We need to work out how to make it work in VS

mlauss2:
> This looks good, thanks. We need to work out how to make it work in VS

A quick google search suggests that VS does support CMake, so the obvious suggestion is to just move the windows project to cmake, no?

ifeldshteyn:
> > This looks good, thanks. We need to work out how to make it work in VS
> 
> A quick google search suggests that VS does support CMake, so the obvious suggestion is to just move the windows project to cmake, no?

What would be the lift to move everything to cmake? At the moment everything in NT is done via vscode builds. If this is a simple procedure we can def move to it but I want to really test it first.

mlauss2:
> > > This looks good, thanks. We need to work out how to make it work in VS
> > 
> > 
> > A quick google search suggests that VS does support CMake, so the obvious suggestion is to just move the windows project to cmake, no?
> 
> What would be the lift to move everything to cmake? At the moment everything in NT is done via vscode builds. If this is a simple procedure we can def move to it but I want to really test it first.

From a cursory glance, maybe this is already enough:  the default cmake build process that works on OSX and linux should also work on windows, just move the if(UNIX) in the top CMakeFile away.  All the other CMakeFiles already have the "if(WIN32)" statements to include the windows-specific files (filesystem, crashhandler).

```diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8df12525..86560db7 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if(ENABLE_TFE)
                set_target_properties(tfe PROPERTIES OUTPUT_NAME "theforceengine")
        endif()
 
-       if(UNIX)
+
                set(THREADS_PREFER_PTHREAD_FLAG ON)
                find_package(Threads REQUIRED)
                find_package(PkgConfig REQUIRED)
@@ -85,7 +85,7 @@ if(ENABLE_TFE)
                )
 
                target_link_directories(tfe PRIVATE ${SDL2_LIBRARY_DIRS} ${SDL2_IMAGE_LIBRARY_DIRS})
-
+       if(UNIX)
                # set up build directory to be able to run TFE immediately: symlink
                # the necessary support file directories into the build env.
                execute_process(COMMAND ln -sf ${CMAKE_SOURCE_DIR}/TheForceEngine/Captions)
@@ -122,14 +122,12 @@ if(ENABLE_TFE)
                set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBUILD_EDITOR")
        endif()
        if(ENABLE_OGV_CUTSCENES)
-               if(UNIX)
                        pkg_check_modules(THEORA REQUIRED theoradec)
                        pkg_check_modules(OGG REQUIRED ogg)
                        pkg_check_modules(VORBIS REQUIRED vorbis vorbisfile)
                        target_include_directories(tfe PRIVATE ${THEORA_INCLUDE_DIRS} ${OGG_INCLUDE_DIRS} ${VORBIS_INCLUDE_DIRS})
                        target_link_libraries(tfe PRIVATE ${THEORA_LIBRARIES} ${OGG_LIBRARIES} ${VORBIS_LIBRARIES})
                        target_link_directories(tfe PRIVATE ${THEORA_LIBRARY_DIRS} ${OGG_LIBRARY_DIRS} ${VORBIS_LIBRARY_DIRS})
-               endif()
                add_definitions("-DENABLE_OGV_CUTSCENES")
        endif()
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBUILD_FORCE_SCRIPT")

```

EDIT: also with a cmake-based windows build, the in-tree SDL2 source+binaries could be removed and fetched and built during the windows build as well, which would make e.g. windows-arm64 builds easier.

PR #621, opened 2026-07-09
Custom projectiles & effects

jerethk:
This PR allows modders to add projectiles and effects beyond DF's enumerated list.
Custom projectiles & effects are read from `json` files (same as custom logics).
Their data are stored in vectors, and they are distinguished from the original projectiles/effects by starting their numbering at 100.

Related changes
- player weapon projectiles are externalised to `weapons.json`
- player weapon projectile spawn offsets are externalised to `weapons.json` instead of being hardcoded
- NPCs will decide to use aimarcing if they have any kind of arcing projectile (not just TDs and mortars)
- 3DO projectiles can be given a flag which disables framerate based scaling

note- this feature has been quite thoroughly tested in the DF-21 community by ben

### DIFF
--- a/TheForceEngine/TFE_DarkForces/hitEffect.cpp
+++ b/TheForceEngine/TFE_DarkForces/hitEffect.cpp
@@ -11,6 +11,7 @@
 #include <TFE_Jedi/Memory/allocator.h>
 #include <TFE_Jedi/Task/task.h>
 #include <TFE_ExternalData/weaponExternal.h>
+#include <TFE_ExternalData/customEffect.h>
 
 using namespace TFE_Jedi;
 
@@ -37,6 +38,8 @@ namespace TFE_DarkForces
 	static EffectData s_effectData[HEFFECT_COUNT];
 	static Allocator* s_hitEffects;
 
+	std::vector<EffectData> s_customEffectData;	// TFE - custom effects
+
 	Task* s_hitEffectTask = nullptr;
 	vec3_fixed s_explodePos;
 	EffectData* s_curEffectData = nullptr;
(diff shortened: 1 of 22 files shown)


TheForceEngine/TFE_DarkForces/projectile.cpp
```diff
@@ -51,29 +53,16 @@ namespace TFE_DarkForces
...
-	static WaxFrame*  s_repeaterProj;
-	static Wax*       s_mortarProj;
-	static Wax*       s_missileProj;
-	static Wax*       s_cannonProj;
-	static Wax*       s_probeProj;
-	static Wax*       s_homingMissileProj;
-	static Wax*       s_bobafetBall;
-	static WaxFrame*  s_landmineWpnFrame;
-	static WaxFrame*  s_landmineFrame;
-	static WaxFrame*  s_thermalDetProj;
```
jerethk, line 66: Note - all this stuff is cleanup for when projectile data was originally externalised. These variables are no longer used.

TheForceEngine/TFE_DarkForces/hitEffect.cpp
```diff
@@ -218,13 +245,13 @@ namespace TFE_DarkForces
 							// obj_addToRefList(obj, ObjRefType_Effect);	// scripting
 						}
 					}
-					if (s_curEffectData->soundEffect)
+					if (s_curEffectData && s_curEffectData->soundEffect)
```
ifeldshteyn, line 248: minor nitpick - can this be a single if (s_curEffectData) clause around and then keep the original (if (s_curEffectData->soundEffect) .... etc...

ifeldshteyn (approved):
Looks good - this should really expand weapon logics.

TheForceEngine/TFE_DarkForces/hitEffect.cpp
jerethk: In terms of minimising diffs, I think this is simpler.
When you indent a whole block, all those lines become diffs
But this way only the 1st line is recorded as a diff

PR #622, opened 2026-07-10
Don't open the menu while escaping a cutscene.

ifeldshteyn:
This allows you to press Escape during a cutscene but not bring up the escape menu.


jerethk (approved):
All good.
(What is going on with your IDE? It made spurious changes to 15 lines - there has to be some setting you have turned on...)

ifeldshteyn:
> All good. (What is going on with your IDE? It made spurious changes to 15 lines - there has to be some setting you have turned on...)

Yeah there must be some weird setting I have yet to find. But I do agree with cleaning up the lines that have spurious spaces when they should be blank.

PR #623, opened 2026-07-10
RFC: CMake improvements for windows builds

mlauss2:
To make cmake build possible on windows:

- use cmake to fetch and build SDL2, SDL2_image if necessary
- use cmake to fetch and build ogg, vorbis if necessary.
  - for theora, because it does not support cmake, use vcpkg on visual studio to fetch it, or rely on availability of the host library on linux/osx.

I've tested this on linux and works pretty well:  If SDL2/SDL2 image are available on the host, they are dynamically linked, otherwise as part of the build step, they are built and statically linked into the executable.  Same for Ogg and Vorbis.
Theora however is difficult. It does not support cmake, just Makefiles and an old visual studio .dsp file.
vcpkg could be used to have it fetched and built before the main application link step.  That is what I opted for here.

Lacking a windows machine atm, I haven't tried to build this with visual studio though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
-cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.24 FATAL_ERROR)
 include(CheckCXXCompilerFlag)
+include(FetchContent)
 
 set(CMAKE_DISABLE_SOURCE_CHANGES ON)
 set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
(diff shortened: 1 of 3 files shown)


mlauss2:
So I got VS to run the build, but it stops at paths.cpp since CMake for some reason configures windows builds for wide chars (i.e. 'char *' is a LPWSTR instead of LPSTR), and I've hit a wall there.
Apart from this it looks fine: sdl2/sdl2image/libremidi/ogg/vorbis all built successfully.

ifeldshteyn:
> So I got VS to run the build, but it stops at paths.cpp since CMake for some reason configures windows builds for wide chars (i.e. 'char *' is a LPWSTR instead of LPSTR), and I've hit a wall there. Apart from this it looks fine: sdl2/sdl2image/libremidi/ogg/vorbis all built successfully.

The encodings are handled in TFE_System/utf8.cpp - please take a look there. I do have bespoke fixes for certain characters (Ex: trademark symbol) and it should be possible to make fixes for widechars if necessary.

mlauss2:
I got it to build now, since some of the Windows api calls were already using the A variant, I just changed all invocations to A.  Now I need to resolve a link failure and it's good to go.

PR #626, opened 2026-07-14
Adding POSIX requirement for building on MacOS

badgeromen:
TO get it to build in MACOS I had to add a POSIX requirment for __APPLE__, it built after this just fine. Before It would not finish the build.


jerethk (approved):
Thanks

PR #627, opened 2026-07-16
Enhance the Saving mechanism.

ifeldshteyn:
Clearly specify the LOAD vs SAVE menu by using distinct colors. 
Alternate the quicksaves so you don't accidentally overwrite them.

image

image

### DIFF
--- a/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
+++ b/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
@@ -1708,21 +1708,36 @@ namespace TFE_FrontEndUI
 	{
 		if (s_saveLoadSetupRequired)
 		{
-			configSaveLoadBegin(save);
+			configSaveLoadBegin(save); 
 		}
 
 		// Create the current display info to adjust menu sizes.
 		DisplayInfo displayInfo;
 		TFE_RenderBackend::getDisplayInfo(&displayInfo);
-
+				
 		f32 leftColumn = displayInfo.width < 1200 ? 196.0f*s_uiScale : 256.0f*s_uiScale;
 		f32 rightColumn = leftColumn + ((f32)TFE_SaveSystem::SAVE_IMAGE_WIDTH + 32.0f)*s_uiScale;
 		const s32 listOffset = save ? 1 : 0;
+
+		string saveLoadText = save ? "Save" : "Load";
+		ImVec4 color = save ? ImVec4(1.0f, 0.65f, 0.0f, 1.0f) : ImVec4(0.0f, 1.0f, 0.0f, 1.0f);
+
+		// Wrap the color around the save/load menu.
+		ImGui::PushStyleColor(ImGuiCol_Border, color);
+		ImGui::PushStyleVar(ImGuiStyleVar_ChildBorderSize, 5.0f);
 
 		// Left Column
 		ImGui::SetNextWindowPos(ImVec2(leftColumn, floorf(displayInfo.height * 0.25f)));
 		ImGui::BeginChild("##ImageAndInfo");
 		{
+			//ImGui::SetCursorPos(ImVec2(leftColumn, floorf(displayInfo.height * 0.20f)));
+			ImGui::PushFont(s_titleFont);
+			ImGui::PushStyleColor(ImGuiCol_Text, color);
+			ImGui::TextWrapped(saveLoadText.c_str());
+			ImGui::PopStyleColor();
+			ImGui::PopFont();
+			ImGui::Spacing();
+
 			// Image
 			ImVec2 size((f32)TFE_SaveSystem::SAVE_IMAGE_WIDTH, (f32)TFE_SaveSystem::SAVE_IMAGE_HEIGHT);
 			size.x *= s_uiScale;
@@ -1764,6 +1779,8 @@ namespace TFE_FrontEndUI
 			ImGui::EndChild();
 		}
 		ImGui::EndChild();
+		ImGui::PopStyleVar();
+		ImGui::PopStyleColor();
 
 		// Right Column
 		f32 rwidth  = 1024.0f * s_uiScale;
--- a/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
+++ b/TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
@@ -1708,15 +1708,15 @@ namespace TFE_FrontEndUI
 	{
 		if (s_saveLoadSetupRequired)
 		{
-			configSaveLoadBegin(save); 
+			configSaveLoadBegin(save);
 		}
 
 		// Create the current display info to adjust menu sizes.
 		DisplayInfo displayInfo;
 		TFE_RenderBackend::getDisplayInfo(&displayInfo);
-				
-		f32 leftColumn = displayInfo.width < 1200 ? 196.0f*s_uiScale : 256.0f*s_uiScale;
-		f32 rightColumn = leftColumn + ((f32)TFE_SaveSystem::SAVE_IMAGE_WIDTH + 32.0f)*s_uiScale;
+
+		f32 leftColumn = displayInfo.width < 1200 ? 196.0f*s_uiScale : 228.0f*s_uiScale;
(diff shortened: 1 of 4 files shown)


jerethk:
Thanks for this. I'll take a closer look when I'm back on my PC.

Just on a quick scan, I think the way it's currently written will cause existing quicksaves to be "lost" to the quickload hotkey. I think an easy way to address this would be to rename slot 1's filename to `quicksave` instead of `quicksave1`. I think that alone should ensure backward compatibility, and everything else should still work as intended.

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
```diff
@@ -1708,21 +1708,36 @@ namespace TFE_FrontEndUI
...
 		{
-			configSaveLoadBegin(save);
+			configSaveLoadBegin(save); 
 		}
 
 		// Create the current display info to adjust menu sizes.
 		DisplayInfo displayInfo;
 		TFE_RenderBackend::getDisplayInfo(&displayInfo);
-
+
```
jerethk: This diff (and the one on line 1711) are detrimental (adding random whitespace), would you mind removing pls?

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
```diff
@@ -1708,21 +1708,36 @@ namespace TFE_FrontEndUI
...
 		// Create the current display info to adjust menu sizes.
 		DisplayInfo displayInfo;
 		TFE_RenderBackend::getDisplayInfo(&displayInfo);
-
+				
 		f32 leftColumn = displayInfo.width < 1200 ? 196.0f*s_uiScale : 256.0f*s_uiScale;
 		f32 rightColumn = leftColumn + ((f32)TFE_SaveSystem::SAVE_IMAGE_WIDTH + 32.0f)*s_uiScale;
 		const s32 listOffset = save ? 1 : 0;
 
+		string saveLoadText = save ? "Save" : "Load";
```
jerethk, line 1764: Can I suggest a few improvements?
- I think giant text is rather obnoxious; I believe the colours are enough to grab attention
- Similar with a super thick border; it doesn't need to be 5pt
- Can we also take this opportunity to reposition the elements better - make them horizontally centred and remove the large gap above

Image

To achieve this, here are my diffs:

```
1718 
-		f32 leftColumn = displayInfo.width < 1200 ? 196.0f*s_uiScale : 256.0f*s_uiScale;
+		f32 leftColumn = displayInfo.width < 1200 ? 196.0f*s_uiScale : 228.0f*s_uiScale;
 
1719
-		f32 rightColumn = leftColumn + ((f32)TFE_SaveSystem::SAVE_IMAGE_WIDTH + 32.0f)*s_uiScale;
+		f32 rightColumn = leftColumn + ((f32)TFE_SaveSystem::SAVE_IMAGE_WIDTH + 64.0f)*s_uiScale;


1727
-		ImGui::PushStyleVar(ImGuiStyleVar_ChildBorderSize, 5.0f);
+		ImGui::PushStyleVar(ImGuiStyleVar_ChildBorderSize, 2.0f);

1730
-		ImGui::SetNextWindowPos(ImVec2(leftColumn, floorf(displayInfo.height * 0.25f)));
+		ImGui::SetNextWindowPos(ImVec2(leftColumn, floorf(displayInfo.height * 0.08f)));

1734
-			ImGui::PushFont(s_titleFont);
+			ImGui::PushFont(s_dialogFont);
```

Can we also get an artist opinion please, @Ben-Dansie ?

TheForceEngine/TFE_Game/saveSystem.cpp
```diff
@@ -44,6 +44,44 @@ namespace TFE_SaveSystem
 	static u32* s_imageBuffer[2] = { nullptr, nullptr };
 	static size_t s_imageBufferSize[2] = { 0 };
 
+	// Two alternating quicksave slots (Quicksave 1 / Quicksave 2), so a
+	// bad quicksave never overwrites your only recent save.
+	static const char* c_quickSaveSlotNames[2] = { "quicksave1.tfe", "quicksave2.tfe" };
```
jerethk: As per my previous comment - you should be able to preserve backward compatibility by changing the first slotname to `quicksave.tfe`

jerethk: Also - please take a look at `c_quickSaveName` in `saveSystem.h`, which is still referenced by `getSaveFilenameFromIndex()` in `saveSystem.cpp` (line 365)

It looks like some backward compatibility is going to be broken here

TheForceEngine/TFE_Game/saveSystem.cpp
```diff
@@ -44,6 +44,44 @@ namespace TFE_SaveSystem
...
+		// we are doing all these comparisons
+		bool oldest = FileUtil::getModifiedTime(path[0]) < FileUtil::getModifiedTime(path[1]);
+		if (getLatest)
+		{
+			return oldest ? 1 : 0;
+		}
+		else
+		{
+			return oldest ? 0 : 1;
+		}
```
jerethk, line 81: I found this hard to read, because "oldest" isn't a good name for a bool
It would be better to name it `slot0IsOlder` or `isSlot0Older`

jerethk (changes requested):

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
jerethk: I have pushed up a change, following on from the discussion in discord. Back over to you :-)

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
```diff
@@ -1614,6 +1615,46 @@ namespace TFE_FrontEndUI
...
+
+		std::sort(s_saveDir.begin() + startIndex, s_saveDir.end(),
+			[](const TFE_SaveSystem::SaveHeader& a, const TFE_SaveSystem::SaveHeader& b) -> bool
+			{
+				if (s_saveSortMode == SAVE_SORT_TIME)
+				{
+					s64 keyA = 0, keyB = 0;
+					TFE_System::parseDateTimeString(a.dateTime, &keyA);
+					TFE_System::parseDateTimeString(b.dateTime, &keyB);
+					return s_saveSortReverse ? keyA < keyB : keyB < keyA;
```
jerethk, line 1646: Much better alternative:

```
time_t timeA, timeB;
timeA = TFE_System::getTimeFromString(a.dateTime);
timeB = TFE_System::getTimeFromString(b.dateTime);
return s_saveSortReverse ? timeA < timeB : timeB < timeA;
```

```
#include <sstream>
#include <iomanip>

time_t getTimeFromString(const char* str)
{
	std::istringstream s(str);
	std::tm timeStruct = {};
	s >> std::get_time(&timeStruct, "%a %b %d %H:%M:%S %Y");
	return std::mktime(&timeStruct);
}
```

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
```diff
@@ -1782,6 +1833,29 @@ namespace TFE_FrontEndUI
 			ImVec2 buttonSize(rwidth - 2.0f, floorf(20 * s_uiScale + 0.5f));
 			ImGui::PushFont(s_dialogFont);
 			ImGui::PushStyleColor(ImGuiCol_Header, ImVec4(0.0f, 0.0f, 0.0f, 0.0f));
+
+			// Sort controls.
+			{
+				const s32  prevSortMode = s_saveSortMode;
+				const bool prevSortReverse = s_saveSortReverse;
+
+				if (ImGui::RadioButton("Name", s_saveSortMode == SAVE_SORT_NAME)) { s_saveSortMode = SAVE_SORT_NAME; }
```
jerethk, line 1842: Please put time first so it is the default. I expect that is what majority of players prefer

jerethk (changes requested):

TheForceEngine/TFE_FrontEndUI/frontEndUi.cpp
jerethk: Have made this change

jerethk (approved):
This is a good enhancement, thanks

PR #629, opened 2026-07-22
Move s_playerDying from pickup to player

jerethk:
It must have gone astray some time during RE


ifeldshteyn (approved):
Thats silly. Correct move.

PR #630, opened 2026-07-25
Create a new "Enhanced Actor" module containing an enhancedAttackFunc

jerethk:
Create `actorEnhanced.cpp`, containing a new `enhancedAttackFunc`
Custom logics will use `enhancedAttackFunc` from now on.
This is where burst fire and other new functionality will exist, allowing the `defaultAttackFunc` to be preserved in its (mostly) original, vanilla form. This will reduce the risk of breaking vanilla NPC functionality in the future, as new functionality is added.

The same approach can be taken for an `enhancedThinkerFunc` and so on.

The `defaultAttackFunc` as it currently stands was copied over to `enhancedAttackFunc`, and then reverted back to its state at commit 7c52d9066cef05b35e555c9f85f595fb147d0331.
(It was not reverted all the way back to vanilla, to avoid breaking old savegames)


TheForceEngine/TFE_DarkForces/Actor/actorSerialization.cpp
```diff
@@ -305,6 +306,31 @@ namespace TFE_DarkForces
 		SERIALIZE(ObjState_BurstFire, attackMod->burstFire.interval, 29);
 		SERIALIZE(ObjState_BurstFire, attackMod->burstFire.shotCount, 5);
 		SERIALIZE(ObjState_BurstFire, attackMod->burstFire.lastShot, 0);
+
+		// Fixup for savegames when burstfire was still in defaultAttackFunc 
+		if (serialization_getMode() == SMODE_READ)
```
ifeldshteyn, line 311: Do we need to check serialization version for this or are you depending on the older saves having hasBurstFire call alone? Have you tested it with older saves?

jerethk: Yes, have tested with older saves.
Serialisation version won't give us enough information. All NPCs in an older save file will have the `defaultAttackFunc`. Because I've moved burstfire to `enhancedAttackFunc`, an NPC which had burstfire will lose it when they deserialise with `defaultAttackFunc`. So they need to be upgraded now to the new `enhancedAttackFunc`. 

Whereas an ordinary vanilla stormtrooper won't have burstfire, so they should still deserialise with `defaultAttackFunc` - they should not be upgraded to enhanced.

ifeldshteyn (approved):
Approved - based on serialization test successes.

PR #631, opened 2026-07-28
Integrated card warn

ifeldshteyn:
Throws a warning message if your GPU is integrated (you are using laptop intel instead of nvidia/amd gpu).  The warning is linux-compatible. 

image

Without this the game may have graphical errors.

image

### DIFF
--- a/TheForceEngine/TFE_RenderBackend/Win32OpenGL/renderBackend.cpp
+++ b/TheForceEngine/TFE_RenderBackend/Win32OpenGL/renderBackend.cpp
@@ -120,8 +120,7 @@ namespace TFE_RenderBackend
 
 		// List of keywords that indicate integrated graphics.
 		// This is not an exhaustive list , but covers the most common integrated graphics.
-		const std::vector<std::string> integratedKeywords = { 
-			"nvidia",
+		const std::vector<std::string> integratedKeywords = { 			
 			"intel(r) uhd",
 			"intel(r) hd",
 			"intel(r) iris",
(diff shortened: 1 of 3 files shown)


TheForceEngine/TFE_RenderBackend/Win32OpenGL/renderBackend.cpp
```diff
@@ -78,6 +78,86 @@ namespace TFE_RenderBackend
 	static GLuint s_globalVAO = 0;
 	static bool s_isMacOS = false;
 
+
+
+	static void showGpuWarningDialog(const char* rendererName)
+	{
+		// Don't show the warning if muted - this is a permanent change unless settings are wiped.
```
jerethk, line 85: Are you totally sure you want to write a setting that someone can never undo without manually editing the settings file?
(I think it is probably ok, but worth being sure.)

jerethk (approved):

TheForceEngine/TFE_RenderBackend/Win32OpenGL/renderBackend.cpp
ifeldshteyn: Yeah i thought about it. But this should only happen once per machine. If you dont care then you can mute it. Its like first start we get a popup about modern or classic. The key is to let user know there is an issue with the machine video settings ans tfe.

mlauss2:
Please revert or at least rework this:  There's nothing inherently "discrete-gpu-only" in the gl renderer at all; the screenshot looks like a windows driver bug.  I've been developing the tfe linux port on an integrated intel haswell gpu, which was introduced 7 years before the rtx3070 in the screenshot, and haven't seen any gfx issues on integrated intel, amd or the quadro k5100m in my old laptop.

jerethk:
> Please revert or at least rework this: There's nothing inherently "discrete-gpu-only" in the gl renderer at all; the screenshot looks like a windows driver bug. I've been developing the tfe linux port on an integrated intel haswell gpu, which was introduced 7 years before the rtx3070 in the screenshot, and haven't seen any gfx issues on integrated intel, amd or the quadro k5100m in my old laptop.

@Karjala22 please note

PR #634, opened 2026-08-10
Undo / redo for level INF data

acpushbutton:
Undo / redo for level INF data

Details of the changes are in each commit, grouped into reasonable chunks.

It's a bit convoluted due to the way INF is stored in memory inside of a Editor_LevelInf.

INF delete not yet implemented.

Resubmitted with commit 5 to address some things.

### DIFF
--- a/TheForceEngine/TFE_Editor/LevelEditor/editCommon.cpp
+++ b/TheForceEngine/TFE_Editor/LevelEditor/editCommon.cpp
@@ -44,7 +44,8 @@ namespace LevelEditor
 				if (sector)
 				{
 					selection_clearHovered();
-					edit_deleteVertex(sector->id, featureIndex, LName_DeleteVertex);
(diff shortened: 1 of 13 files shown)


jerethk:
Hi @acpushbutton thanks for this. It's a big one so let's go through it piece by piece, I'll do my best as time permits.

Can I start by asking- is this just for INF edits, or is there also changes to the geometry (wall/vertex) editor?

acpushbutton:
Apologies @jerethk . 1597ca0 is not supposed to be here. I didn't realise that appending to this on my branch would auto-update the PR to the main project too.

The submission is intended to be from 10d1df6 to 7ca8333.

I'll take your advice on how best to fix this.

acpushbutton:
Closing this, I'll fix it up and resubmit another day.

jerethk:
@acpushbutton no worries at all. Yeah if they were supposed to be 2 separate additions, I think it would be best to have them separated into 2 PRs. Thanks for that.

PR #635, opened 2026-08-16
Enhancements for selection_buildDerived()

acpushbutton:
This function handles converting selections of one type to another. Like sectors to walls / vertexes. This updates it to complete all conversion types, namely from simpler ones like verts to walls / sectors.

### DIFF
--- a/TheForceEngine/TFE_Editor/LevelEditor/selection.cpp
+++ b/TheForceEngine/TFE_Editor/LevelEditor/selection.cpp
@@ -6,6 +6,8 @@
 #include "sharedState.h"
 #include "infoPanel.h"
 
+#include <unordered_map>
+
 using namespace TFE_Editor;
 
 namespace LevelEditor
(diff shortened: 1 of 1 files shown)


TheForceEngine/TFE_Editor/LevelEditor/selection.cpp
```diff
@@ -745,26 +747,114 @@ namespace LevelEditor
...
-			// For now, vertices don't select anything.
-			// This may change.
+			const s32 count = (s32)s_selectionList2[SEL_VERTEX].size();
+			const FeatureId* vtx = s_selectionList2[SEL_VERTEX].data();
+
+			s32 featureIndex = -1;
+			std::unordered_map<s32, std::vector<s32>> sectorVtxMap;
+
+			// Fill the hashmap with sectors and feature (vtx) ids.
+			for (size_t i = 0; i < count; i++)
```
jerethk: Minor: other loops in this method use `s32/u32` instead of `size_t`
The code is a bit tidier and easier to read if it's consistent

Same with the other loops further down

(I think it is desirable to stick with TFE's type system _as much as possible_, though I can see there are a few places in this file where even Lucius has used `size_t`)

Also, you have declared `count` as a `s32` so `i` should be signed too

TheForceEngine/TFE_Editor/LevelEditor/selection.cpp
```diff
@@ -745,26 +747,114 @@ namespace LevelEditor
...
+									if (vtxIds[v1] == w->idx[1])
+									{
+										FeatureId id = createFeatureId(sector, (s32)wallIndex, HP_MID);
+										selection_insertFeatureId(s_selectionList2[SEL_SURFACE], id);
+									}
+								}
+							}
+						}
+					}
+
```
jerethk: Unnecessary blank line

TheForceEngine/TFE_Editor/LevelEditor/selection.cpp
```diff
@@ -745,26 +747,114 @@ namespace LevelEditor
...
 			for (s32 i = 0; i < count; i++)
 			{
 				s32 featureIndex;
 				HitPart part;
 				EditorSector* sector = unpackFeatureId(list[i], &featureIndex, (s32*)&part);
+
 				// Skip flats.
 				if (part == HP_FLOOR || part == HP_CEIL) { continue; }
 
+				// Populate hashmap for sector selection based on MID walls
```
jerethk: Your comment says you are basing on MID, but I can't see you filtering to MID anywhere below?

Same in the section below (from :845)

TheForceEngine/TFE_Editor/LevelEditor/selection.cpp
```diff
@@ -776,6 +866,7 @@ namespace LevelEditor
 				FeatureId id = createFeatureId(sector);
 				selection_insertFeatureId(s_selectionList2[SEL_SECTOR], id);
 			}
+
```
jerethk: unnecessary blank line

TheForceEngine/TFE_Editor/LevelEditor/selection.cpp
```diff
@@ -745,26 +747,114 @@ namespace LevelEditor
...
+			const FeatureId* vtx = s_selectionList2[SEL_VERTEX].data();
+
+			s32 featureIndex = -1;
+			std::unordered_map<s32, std::vector<s32>> sectorVtxMap;
+
+			// Fill the hashmap with sectors and feature (vtx) ids.
+			for (size_t i = 0; i < count; i++)
+			{
+				EditorSector* sector = unpackFeatureId(vtx[i], &featureIndex);
+				if (sectorVtxMap.count(sector->id) > 0)
```
jerethk, line 761: We need a `nullptr` check here, because `unpackFeatureId` can return `nullptr`

eg. `if (!sector) { continue; }`

jerethk (changes requested):
Hi, thanks for this - just a couple things to look at, which are relatively simple things.
The MID issue looks like it is potentially a bug

BTW - I haven't tested it, and I probably won't get round to it - so I'm relying on you having tested it all works as expected :-)

acpushbutton:
Changes made see 2464899.

Re: size_t, I tend to use this when the var is used for array indexing only. Since other methods like vec.size() return size_t it means I can avoid some casts. Mixing unsigned and signed was definitely an error though.

Re: MID walls, the logic was 98% correct but the comments were wrong.

TheForceEngine/TFE_Editor/LevelEditor/selection.cpp
```diff
@@ -842,7 +841,7 @@ namespace LevelEditor
 				selection_insertWallVertices(sector, wall);
 			}
 
-			// Do select sectors on MID wall tests
+			// Select whole sectors if all walls are accounted for
```
jerethk: Sorry can you please clarify for me -- is the logic here that a wall is considered "accounted for" if ANY one of its parts (bot, mid or top) is in the selection?
i.e. there is no need for both bot and top to be included?

jerethk: Something I am wondering. Sectors can be huge/complex, with dozens of walls.
If a sector has 40 walls, do you want to require that the user selects all 40 for it to be recognised and converted into the sector selection? 

I wonder if it would be better if the logic was: if ANY wall is selected (even just 1), its sector will become part of the selection.

Especially considering Lucius' logic from line 857 onwards: if only the floor _or_ ceiling is selected, the whole sector becomes selected when changing from surface to sector mode

TheForceEngine/TFE_Editor/LevelEditor/selection.cpp
jerethk: Is this an "and" or an "or"?
Does the user need to have both floor and ceil selected for it to count?

jerethk:
> Re: size_t, I tend to use this when the var is used for array indexing only. Since other methods like vec.size() return size_t it means I can avoid some casts. Mixing unsigned and signed was definitely an error though.
> 

Oh, thanks for explaining this. I didn't realise that switching to `s32` would force you to have to do those typecasts. I agree they would be nice to avoid. If you wish, feel free to change back to `size_t` to avoid the typecasts.

TheForceEngine/TFE_Editor/LevelEditor/selection.cpp
jerethk: NB. Has been discussed in Discord, and I'm more than happy to leave it as-is, if this is the desired behaviour.

acpushbutton:
Final commit f31c54d - Fixes up the logic to ensure all wallIds are accounted for when transiently selecting their parent sector.

jerethk (approved):
Great work

PR #642, opened 2026-09-19
Fix for empty GOL file for Dark Trooper Center

ifeldshteyn:
Fix for empty GOL file to match DOS.

image


TheForceEngine/TFE_Jedi/Level/level.cpp
```diff
@@ -638,6 +638,12 @@ namespace TFE_Jedi
 		}
 
 		size_t len = file.getSize();
+		if (len == 0)
+		{
+			TFE_System::logWrite(LOG_ERROR, "level_loadGoals", "Level goals file '%s' is empty.", levelName);
+			return JFALSE;
```
jerethk, line 644: You need to close the file (stream) before quitting out of this function, otherwise there will probably be a resource leak.

jerethk (changes requested):

TheForceEngine/TFE_Jedi/Level/level.cpp
jerethk: Ok, so it turns out that `Filestream` has a destructor that calls `close()` when an open Filestream moves out of scope.

Even so, I would _personally_ prefer it if you could move the early return to after the `file.close()` because then the code here would read more as if it is doing the right thing. Everywhere else it opens files, TFE explicitly closes them with a `file.close()` inline in the code (LEV, O, INF, VUEs, textures, 3DOs etc.) so I think that it is best not to depart from the pattern here, purely as a matter of consistency.

The destructor is still there as a "backstop", but we don't normally rely on it.

jerethk (approved):
Thanks!

PR #643, opened 2026-09-19
Mute LevelScript errors. 

ifeldshteyn:
Don't throw errors if Level Script is missing. Only throw errors if you cannot open it. This will mute the error spam for all mods.


TheForceEngine/TFE_ForceScript/Angelscript/add_on/scriptbuilder/scriptbuilder.cpp
```diff
@@ -190,14 +190,21 @@ s32 readScriptFileTFE(const char* filename, string& code, asIScriptEngine* engin
 {
 	FilePath path;
 	FileStream file;
-	if (!TFE_Paths::getFilePath(filename, &path) || !file.open(&path, FileStream::MODE_READ))
+	if (!TFE_Paths::getFilePath(filename, &path))
```
jerethk, line 193: @ifeldshteyn I agree that having regular error messages about missing levelscript.fs is annoying.

However, this is also used for other `.fs` files that may be `#include`d along with levelscript.fs
For example if levelscript.fs has `#include secbase.fs`
and `secbase.fs` is absent, we DO want an error message to let us know this.

Perhaps the solution is an extra `if` within this block

```
if (filename != c_levelScriptFile)
{
   string msg = "Script file not found: '" + std::string(filename) + "'";
   engine->WriteMessage ....
}
```

jerethk (changes requested):

PR #644, opened 2026-09-19
Correct INF Trigger Event field

acpushbutton:
Event field in Trigger data incorrectly points to eventMask.
- Update field to point to event instead
- Change ImGui widget to read unsigned int.
- Update checkbox bit flipper to point to event.
eventDatafix


jerethk (approved):
Good catch!

